PR #25, opened 2026-07-06
Skip cross-repo round-trip when the sibling buyer repo is absent (fresh clone pytest fails)

aleksUIX:
## Summary

A fresh `git clone` + `pip install -e ".[dev]"` + `pytest tests/unit` fails out of the box:

```
E  RuntimeError: Could not locate ad_seller_system in path ancestry of
   .../seller-agent/tests/unit/test_openrtb_parser.py; set AD_BUYER_SRC_PATH to override.
tests/unit/test_openrtb_parser.py:250
```

The cross-repo round-trip test walks parent directories looking for one literally named `ad_seller_system` (a default clone is named `seller-agent`), then needs a sibling `ad_buyer_system` checkout. CI stays green only because `ci.yml` checks out buyer-agent and sets `AD_BUYER_SRC_PATH`; every standalone contributor clone gets a red suite.

buyer-agent hit the mirror image of this and fixed it in buyer-agent#92 ("Skip cross-repo audience plan round-trip when sibling repos absent"). This PR ports that exact pattern:

- skip (with the same explanatory message style) instead of raising when the seller repo root cannot be located
- skip when the resolved buyer `src/` directory does not exist (partial checkout)

## Verification

- Standalone clone: the test now reports `SKIPPED`, file exit code 0, 12 others pass.
- Opt-in still works: with `AD_BUYER_SRC_PATH` pointing at a real buyer-agent checkout's `src/`, the round-trip runs and passes.
- `ruff check` clean.

### DIFF
--- a/tests/unit/test_openrtb_parser.py
+++ b/tests/unit/test_openrtb_parser.py
@@ -232,6 +232,8 @@ def test_round_trip_builder_then_parser_recovers_refs() -> None:
     import sys
     from pathlib import Path
 
+    import pytest
+
     # Path resolution (per ar-e2rj): tests can override via the
     # `AD_BUYER_SRC_PATH` env var. Otherwise, walk up from this file to
     # find the seller repo root (`ad_seller_system`); the buyer repo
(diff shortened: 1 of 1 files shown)


Sirajmx:
Thanks, @aleksUIX! Merged.

PR #28, opened 2026-07-10
Feature/integrate llm providers

Sirajmx:
Summary

  - Adds OpenAI and Gemini as supported LLM providers alongside the existing Anthropic default
  - Installs the required crewai extras (openai, gemini) in pyproject.toml so dependencies are present when a non-Anthropic model is configured
  - Adds OPENAI_API_KEY and GOOGLE_API_KEY as optional fields in Settings with inline provider-switch docs
  - Updates model references from gemini-2.0-flash (deprecated June 2026) to gemini-2.5-flash across settings.py, .env.example, and
  docs/guides/configuration.md
  - Fixes MANAGER_LLM_MODEL default in .env.example from claude-sonnet to claude-opus-4 to match settings.py

  Provider configuration (.env.example)

  Only one provider can be active at a time. The file ships with Anthropic uncommented as the default. To switch, comment out the Anthropic block
  and uncomment the target block:

  --Provider: Anthropic (default — active)
  ANTHROPIC_API_KEY=your-anthropic-api-key-here
  DEFAULT_LLM_MODEL=anthropic/claude-sonnet-4-5-20250929
  MANAGER_LLM_MODEL=anthropic/claude-opus-4-20250514

  -- Provider: OpenAI (alternative)
  --OPENAI_API_KEY=your-openai-api-key-here
  --DEFAULT_LLM_MODEL=openai/gpt-4o
  -- MANAGER_LLM_MODEL=openai/gpt-4o

  --Provider: Google Gemini (alternative)
  --GOOGLE_API_KEY=your-google-api-key-here
  --DEFAULT_LLM_MODEL=gemini/gemini-2.5-flash
  --MANAGER_LLM_MODEL=gemini/gemini-2.5-flash

  Test plan

  - [ ] Start seller agent with default Anthropic config — agents initialize without error
  - [ ] Swap to openai/gpt-4o + OPENAI_API_KEY — crew runs using OpenAI
  - [ ] Swap to gemini/gemini-2.5-flash + GOOGLE_API_KEY — crew runs using Gemini
  - [ ] Verify pyproject.toml extras install cleanly: pip install -e ".[anthropic,openai,gemini]"

PR #33, opened 2026-07-20
Fix crewai extras: gemini -> google-genai

Kishore-2572:
## Summary
- `crewai[anthropic,bedrock,openai,gemini]` (from PR #28) declares two invalid extras: `openai` isn't a real extra (the `openai` SDK ships as a core CrewAI dependency already, so it's a harmless no-op), and `gemini` isn't the correct extra name — it silently fails to install with a `pip` warning ("does not provide the extra 'gemini'") and never pulls in the `google-genai` package Gemini actually needs.
- Verified directly: `pip install -e . --dry-run` against the old spec produced `WARNING: crewai 1.14.6 does not provide the extra 'gemini'`, and `google-genai` was absent from the resolved package list.
- Fix: `crewai[anthropic,bedrock,google-genai]` — the correct extra name, confirmed present in `crewai`'s own `Provides-Extra` metadata.

## Test plan
- [x] `pip install -e . --dry-run` — exit 0, zero "does not provide the extra" warnings, `google-genai` resolves correctly
- [x] Confirmed `anthropic`, `boto3`, `google.genai`, `openai` packages all importable after install
- [x] Re-ran all 4 LLM provider flows end-to-end (settings resolution → real API call → real agent `kickoff()`) after this change:

| Provider | Model | provider field | Real call | Agent kickoff |
|---|---|---|---|---|
| Anthropic | claude-sonnet-4-5-20250929 | anthropic | ✅ "pong pong pong" | ✅ correct CPM answer |
| OpenAI | gpt-4o | openai | ✅ "pong pong pong" | ✅ correct CPM answer |
| Gemini | gemini-2.5-flash | gemini | ✅ "pong pong pong" | ✅ correct CPM answer |
| Ollama (base_url) | gemma4:latest | openai (forced) | ✅ "pong pong pong" | ✅ correct CPM answer |

Gemini is the meaningful confirmation here — it exercises the actual `google-genai` package this fix installs; the old extras line would have left it missing.

### DIFF
--- a/tests/unit/test_endpoint_no_flow_kickoff.py
+++ b/tests/unit/test_endpoint_no_flow_kickoff.py
@@ -116,7 +116,7 @@ async def test_list_products_returns_200_with_products_key(client):
     assert "product_id" in p
     assert "name" in p
     assert "inventory_type" in p
-    assert "basePrice" in p
+    assert "base_cpm" in p
     assert "deal_types" in p
 
 
--- a/tests/unit/test_endpoint_no_flow_kickoff.py
+++ b/tests/unit/test_endpoint_no_flow_kickoff.py
@@ -116,7 +116,7 @@ async def test_list_products_returns_200_with_products_key(client):
     assert "product_id" in p
     assert "name" in p
     assert "inventory_type" in p
-    assert "base_cpm" in p
+    assert "basePrice" in p
     assert "deal_types" in p
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ classifiers = [
 ]
 
 dependencies = [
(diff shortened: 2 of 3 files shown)


tests/unit/test_endpoint_no_flow_kickoff.py
```diff
@@ -116,7 +116,7 @@ async def test_list_products_returns_200_with_products_key(client):
     assert "product_id" in p
     assert "name" in p
     assert "inventory_type" in p
-    assert "base_cpm" in p
+    assert "basePrice" in p
```
therevoltingx: why are there changes that are out of scope?  this seems inconsistent why is there a mixture of snake case and camel case

Kishore-2572: @therevoltingx It seems the changes already in place in main

PR #35, opened 2026-07-26
Replace retired manager LLM default with claude-opus-4-8

aleksUIX:
## Problem

The shipped `MANAGER_LLM_MODEL` default is `anthropic/claude-opus-4-20250514`. Anthropic retired that model on June 15, 2026 ([deprecation schedule](https://platform.claude.com/docs/en/about-claude/models/migration-guide)), so any fresh install with an Anthropic key now fails at runtime with `model_not_found` the first time a manager agent is invoked. `DEFAULT_LLM_MODEL` is unaffected (Sonnet 4.5 is still active).

## Change

Swap the default to `anthropic/claude-opus-4-8`, Anthropic's designated replacement for Opus 4. Updated everywhere the old id was pinned: `settings.py`, `.env.example`, and `docs/guides/configuration.md`.

The commented Bedrock example becomes `bedrock/us.anthropic.claude-opus-4-8`: verified against LiteLLM's model registry, which carries `us.anthropic.claude-opus-4-8` as the current US inference profile id (the `-v1:0` suffix was dropped for undated model ids).

Mirrors the same fix submitted to buyer-agent, keeping the two repos' LLM defaults in sync.

## Verification

- `Settings(anthropic_api_key=...)` instantiates with the new default; no occurrence of the retired id remains in the repo.
- Note: the full suite cannot run on a public clone because `pyproject.toml` pins the private `iab-agentic-primitives` repo (same reason CI on main has been failing since ~July 20); unrelated to this change, filing separately.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -153,7 +153,7 @@ PREFERRED_DEAL_DISCOUNT_MAX=0.15
 # Provider: Anthropic (default)
 ANTHROPIC_API_KEY=your-anthropic-api-key-here
 DEFAULT_LLM_MODEL=anthropic/claude-sonnet-4-5-20250929
-MANAGER_LLM_MODEL=anthropic/claude-opus-4-20250514
+MANAGER_LLM_MODEL=anthropic/claude-opus-4-8
 LLM_TEMPERATURE=0.3
 LLM_MAX_TOKENS=4096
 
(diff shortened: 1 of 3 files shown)


atc964:
Merged — thanks. One note for a follow-up: the commented Bedrock example (bedrock/us.anthropic.claude-opus-4-8) drops the -v1:0 suffix pattern the Sonnet line uses; worth verifying the exact Bedrock inference-profile ID before anyone relies on the comment.

atc964:
Following up on my earlier note: verified against the AWS Bedrock model card and LiteLLM's registry — your ID was correct as written. us.anthropic.claude-opus-4-8 is the real US cross-region inference profile; Opus 4.7+ dropped the dated -v1:0 form, so the asymmetry with the Sonnet line is expected. No change needed.

PR #38, opened 2026-07-27
Seller agent agentic e2e full implementation

ghost:


atc964:
Noticed this was closed without a comment — no problem if it's withdrawn, but a one-line close reason helps us and future readers (superseded? rethinking the approach? found a blocker?). If there's work in here you want to land, happy to review a rescoped version.

Issue #41, opened 2026-08-05
pricing_verified appears to require a prior matching quote; the quote-then-propose sequence is undocumented


jaanijuk:
`pricing_verified` on a proposal response appears to require a matching quote previously issued by the seller before it can return `true`. In the observed flow, it does not verify an unquoted proposal directly against the rate card or floor - which makes it sequence-dependent in a way nothing currently documents, and in the observed flow the intuitive order (propose first) does not verify.

**Minimal reproduction** (seller v2.4.0, `1c4114e`; default catalogue; WSL2 Ubuntu 24.04, Python 3.12.3):

1. Propose with no prior quote →

```
pricing_verified: false
pricing_verification_reason: "No matching quote found for this buyer and product."
```

2. Create a quote **for the same buyer identity and product** (`POST /api/v1/quotes` → `qt-367309d94d20`, $15.00 CPM - the anonymous/public buyer in our run, so anchoring works even unauthenticated), then propose that product **at the quoted price** →

```
status: accepted   pricing_verified: true
pricing_verification_reason: "Proposed CPM $15.00 matches quote qt-367309d94d20 ($15.00, 0.0% difference)."
```

The `false` path we've observed in two environments (WSL2 and a clean Ubuntu 24.04 container); the `true` path on WSL2. The reason strings are excellent, incidentally (precise and causeful) which is what made the mechanism discoverable at all.

## Why this is worth a docs line

As far as we can find, `pricing_verified` appears only in `docs/api/openapi.json`, the quote-then-propose sequence isn't described anywhere in prose. An operator running the natural flow sees `false` on every proposal we observed and reasonably concludes price verification is broken, when it's actually waiting for an anchor. If the anchoring is deliberate and it reads like a sensible anti-hallucination guard, tying an agent's claimed price to a record the seller itself issued - then the sequence is a feature that deserves stating.

## Questions rather than prescriptions

1. Is a pre-existing quote intentionally a prerequisite for `pricing_verified: true`? (We're assuming yes, and that it's a trust-verification design.)
2. If so, could the sequence be documented outside the OpenAPI schema as the README's flow description or the demo docs seem the natural homes?
3. Was a fallback (verifying an unquoted proposal directly against the rate card / floor) considered and deliberately rejected? If quote-anchoring is the stronger guarantee, that reasoning is itself worth a sentence in the docs, so integrators don't "helpfully" request the weaker behaviour later.

[...]

PR #43, opened 2026-08-06
docs: document the quote-then-propose sequence for pricing_verified

jaanijuk:
Documents the quote-then-propose sequence for `pricing_verified`, as invited on #41.

Per the confirmation there: verification is quote-anchored by design, with no rate-card or floor fallback, because an unanchored "verified" would defeat the purpose of the field. Before this change `pricing_verified` appeared only in  `docs/api/openapi.json`, so an operator following the intuitive order (propose first, without a prior matching quote) receives `false` and could reasonably conclude verification is broken.

**Changes**

- `docs/guides/pricing-rules.md` -- new **Price Verification (Quote-Anchored)** section covering the mechanism, required sequence, and outcomes. Placed after the negotiation material and before cross-advertiser consistency, both price-integrity topics.
- `docs/api/overview.md` -- cross-reference under **Proposals**, where the field appears on the response.
- `README.md` -- clarifying phrase on the Proposals endpoint-table row.

**Verified against** seller-agent `v2.4.1`, all three paths:

- No prior quote -> `false`, "No matching quote found for this buyer and product."
- Quote at $15.00, propose at $15.00 -> `true`, "Proposed CPM $15.00 matches quote qt-367309d94d20 ($15.00, 0.0% difference)."
- Quote at $15.00, propose at $18.00 -> `false`, "Proposed CPM $18.00 is outside tolerance of all active quotes for this buyer and product."

Docs only -- no code or test changes. Happy to adjust placement or wording, and to add cross-references elsewhere if useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ INDEX_EXCHANGE_API_URL=https://api.indexexchange.com
 | Deals | 1 | Deal creation from accepted proposals |
 | Deal Performance | 1 | Delivery metrics |
 | Bulk Operations | 1 | Batch deal create/update/cancel |
-| Proposals | 1 | Proposal submission |
+| Proposals | 1 | Proposal submission (quote-anchored price verification) |
 | Negotiation | 3 | Counter-offers + negotiation state and messages |
 | Discovery | 1 | Natural-language inventory discovery |
 | Audience | 1 | Agentic audience matching |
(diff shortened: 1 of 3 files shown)


atc964 (approved):
Accuracy-checked against the verification code path — every reason string and behavioral claim matches, including the endpoint prefixes (which faithfully reflect the actual router registrations). Strict docs build passes. Merging once CI confirms.

Two optional additions if you feel like a follow-up: state the default tolerance explicitly (1% relative difference, DEFAULT_TOLERANCE_PCT, overridable), and add the third false-reason to the outcomes table — "All quotes for this buyer and product have expired." — so the enum reads exhaustively.

PR #47, opened 2026-08-10
fix: chat-mode router misses 'list products' and 'create a deal'

### LINKED ISSUE #46: Chat-mode `list products` falls through to `general`; integration test and smoke check pass on the fallback
## Summary

`list products` - the repository's documented chat-mode example, the default AgentCore smoke-test prompt, and the input to `TestChatMode::test_list_products` - matches no intent in the chat-mode keyword router and falls through to the general handler. Neither automated check detects this: the integration assertion is satisfied by the fallback text, and the deployment smoke check looks only for error markers.

A second prompt shipped with the repository is also misrouted, though less consequentially: the deal example printed by the general handler itself is classified as `availability` rather than `deal`. Both cases stem from literal substring matching against phrasings distributed alongside the matcher.

## Observed: the router

`process_message` lowercases the message and tests five intents in order ([`chat/main.py#L255-L266`](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/src/ad_seller/interfaces/chat/main.py#L255-L266)). Matching is plain substring containment, with no normalisation beyond `.lower()`.

| Intent | Line | Keywords |
|---|---|---|
| deal | [L359](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/src/ad_seller/interfaces/chat/main.py#L359) | `create deal`, `book`, `buy inventory`, `want to buy`, `make a deal` |
| counter-offer | [L369](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/src/ad_seller/interfaces/chat/main.py#L369) | (offer terms) |
| pricing | [L364](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/src/ad_seller/interfaces/chat/main.py#L364) | `price`, `cost`, `cpm`, `rate`, `how much` |
| availability | [L385](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/src/ad_seller/interfaces/chat/main.py#L385) | `available`, `inventory`, `impressions`, `capacity` |
| general | [L643](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/src/ad_seller/interfaces/chat/main.py#L643) | fallback |

## Observed (case 1): `list products`

No keyword set contains `product` or `list`, so the message reaches the general handler and returns `"type": "general"`.

That exact string appears in three places:

- [`docs/guides/agentcore-deployment.md#L62`](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/docs/guides/agentcore-deployment.md#L62) — the documented chat-mode example
- [`infra/aws/agentcore/deploy.sh#L37`](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/infra/aws/agentcore/deploy.sh#L37) — the default `--prompt` used by `--test` and `--test-only`
- [`tests/integration/agentcore/test_runtime.py#L185-L193`](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/tests/integration/agentcore/test_runtime.py#L185-L193) — `TestChatMode::test_list_products`

## Observed (case 2): the advertised deal example

The general handler prints three example questions ([L659-L662](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/src/ad_seller/interfaces/chat/main.py#L659-L662)). The third is:

```
"I want to create a deal for 5M display impressions"
```

The deal matcher looks for `create deal` without an article, so `create a deal` does not match. Nor do `book`, `buy inventory`, `want to buy` or `make a deal`. The message then reaches the availability matcher and hits on `impressions`.

Result: the deal intent is not reached by the example phrasing the handler itself prints, and a deal request returns an inventory listing. Other phrasings do match the deal matcher — `create deal`, `book`, `make a deal` — so this is specific to the advertised wording.

## Observed: why neither check catches case 1

**The integration test.** Its docstring states the expectation plainly:

```python
"""Chat mode responds to 'list products' with inventory data."""
...
response = result["response"].lower()
assert any(kw in response for kw in ["product", "inventory", "ctv", "video", "display"])
```

Run against the actual fallback response, that assertion matches **all five** keywords — `product` from "specific products or ranges", and `inventory`, `ctv`, `video`, `display` from the menu headings and the example questions:

```
intent: general | assertion passes: True | matched: ['product', 'inventory', 'ctv', 'video', 'display']
```

Because the fallback contains every asserted domain term, a content-keyword assertion is the wrong level of contract for the routing behaviour this test is intended to cover.

**The deployment smoke test.** [`deploy.sh#L797-L802`](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/infra/aws/agentcore/deploy.sh#L797-L802) greps the invocation output for `"error":`, `"exception":` or `Invocation failed`, and sets `passed=true` when none are present. The fallback response contains none of them.

## Reproduction

No server, ports or AWS dependencies required:

```python
from ad_seller.interfaces.chat.main import ChatInterface

c = ChatInterface()
for p in ['list products',
          'what CTV inventory do you have?',
          'how much does video inventory cost?',
          'I want to create a deal for 5M display impressions']:
    print(f'{p:46} -> ' + c.process_message(p)['type'])
```

```
list products                                  -> general
what CTV inventory do you have?                -> availability
how much does video inventory cost?            -> pricing
I want to create a deal for 5M display impressions -> availability
```

Rows 2 and 3 are controls, confirming the router functions. Rows 1 and 4 are the two strings it misses.

The same four results were obtained through the AgentCore HTTP entrypoint via `POST /invocations`.

## Environment

seller-agent v2.4.1, commit `2bb7ebce45891e3bdddd7a43198cb81d710b0916`; also present on `main` at the same line numbers as of 8 August 2026. Verified in two environments: an existing checkout, and a clean room built with `uv sync --locked` (the install CI uses).

## Inferred

- **`deploy.sh --test` behaviour on a live runtime.** Read from the script, not executed - I have not run a deployment.
- **Whether this is intended.** The test docstring says chat mode should respond to `list products` with inventory data, which suggests it is not. If chat mode is regarded as a demo path rather than a supported surface, the routing question is a design call for maintainers - but the test and smoke-test behaviour would still merit attention, since both currently report success against a runtime that is misrouting.

## Suggested remedy

Two independent parts, and the second holds regardless of how the first is decided.

**Routing.** Either add catalogue terms to the router so `list products` reaches an intent, or change the prompts shipped in the docs, `deploy.sh` and the test to phrasings that already match. For case 2, `create a deal` matching `create deal` is a one-word fix, but the general pattern (literal substrings that miss the phrasings printed alongside them) may warrant a broader look.

**Test assertion.** The current assertion cannot distinguish a successful route from the fallback. A routing-contract assertion would, but the helper makes that awkward: `invoke_runtime` documents its return as `response: str (the text response)`, so the structured payload carrying the `type` field is flattened before the test sees it. Either preserve the parsed response in the helper and assert on the routed type, or assert against `result["raw"]`, which retains the full invocation output.

## Acceptance criteria

- `list products` no longer returns a `general` response in chat mode, or the documentation, the smoke-test default and the integration-test input are changed together to a prompt the router does match.
- The chat-mode integration test fails when the router falls through to the general handler.
- The advertised deal example either routes to `deal`, or is replaced with a phrasing that does.

garvitkaushik-123:
## Summary

The chat-mode keyword router in `ChatInterface.process_message` falls through to the general handler for two prompts that the repository itself ships as documented examples:

- **`list products`** — the default AgentCore smoke-test prompt, the documented chat-mode example, and the input to `TestChatMode::test_list_products` — matches no intent because `product` is absent from the availability keyword set.
- **`I want to create a deal for 5M display impressions`** — the third example printed by the general handler itself — misses the deal matcher because `create deal` doesn't match `create a deal` (the article is missing from the keyword list). It instead hits the availability matcher on `impressions`.

Neither the integration test nor the deployment smoke check caught this: the integration assertion was satisfied by the general fallback's text (which contains all five asserted keywords), and the smoke check only looks for error markers.

## Changes

**Router keywords** (`src/ad_seller/interfaces/chat/main.py`):
- Add `product` and `catalog` to `_is_availability_inquiry` so `list products` routes to availability
- Add `create a deal` to `_is_deal_request` (before `create deal` so the longer match is tested first) so the general handler's own example prompt reaches the deal handler

**Integration test** (`tests/integration/agentcore/test_runtime.py`):
- `test_list_products` now asserts `"type": "general"` is absent from the raw response, catching routing fallthrough that the content-keyword assertion alone cannot distinguish

**Unit test** (`tests/unit/test_chat_routing.py`):
- Parametric test covering all five intent types with representative prompts
- Guard test verifying the general handler's example prompts do not route back to general

## Test plan

- [x] `pytest tests/unit/test_chat_routing.py -v` — 13 tests pass
- [x] Full unit suite (`pytest tests/unit/`) — 1393 tests pass, no regressions
- [x] `ruff check` and `ruff format --check` clean

Closes #46

### DIFF
--- a/src/ad_seller/interfaces/chat/main.py
+++ b/src/ad_seller/interfaces/chat/main.py
@@ -356,7 +356,14 @@ class ChatInterface:
 
     def _is_deal_request(self, message: str) -> bool:
         """Check if message is a deal creation request."""
-        deal_keywords = ["create deal", "book", "buy inventory", "want to buy", "make a deal"]
+        deal_keywords = [
+            "create a deal",
+            "create deal",
+            "book",
+            "buy inventory",
+            "want to buy",
+            "make a deal",
+        ]
         return any(keyword in message for keyword in deal_keywords)
 
     def _is_pricing_inquiry(self, message: str) -> bool:
@@ -382,7 +389,7 @@ class ChatInterface:
 
     def _is_availability_inquiry(self, message: str) -> bool:
         """Check if message is an availability inquiry."""
-        avail_keywords = ["available", "inventory", "impressions", "capacity"]
+        avail_keywords = ["available", "inventory", "impressions", "capacity", "product", "catalog"]
         return any(keyword in message for keyword in avail_keywords)
 
     def _handle_deal_request(
--- a/tests/integration/agentcore/test_runtime.py
+++ b/tests/integration/agentcore/test_runtime.py
@@ -183,10 +183,13 @@ class TestChatMode:
     """Tests for the chat routing mode (keyword-based ChatInterface)."""
 
     def test_list_products(self, runtime_config):
-        """Chat mode responds to 'list products' with inventory data."""
+        """Chat mode routes 'list products' to availability, not the general fallback."""
         result = invoke_runtime(runtime_config, {"prompt": "list products"})
         assert result["success"], f"Invoke failed: {result['error']}"
-        # Should mention products or inventory
+        raw = result["raw"].lower()
+        assert '"type": "general"' not in raw, (
+            f"'list products' fell through to general handler: {result['response'][:200]}"
+        )
         response = result["response"].lower()
         assert any(kw in response for kw in ["product", "inventory", "ctv", "video", "display"]), (
             f"Response doesn't mention products: {result['response'][:200]}"
--- a/tests/unit/test_chat_routing.py
+++ b/tests/unit/test_chat_routing.py
@@ -1,0 +1,57 @@
+"""Tests for ChatInterface keyword routing.
+
+Verifies that documented and advertised prompts reach the correct intent
+handler rather than falling through to the general fallback.
+"""
+
+import pytest
+
+from ad_seller.interfaces.chat.main import ChatInterface
+
+
+@pytest.fixture
+def chat():
+    return ChatInterface()
+
+
+class TestChatIntentRouting:
+    """Each prompt should map to the expected intent type."""
+
+    @pytest.mark.parametrize(
+        "prompt, expected_type",
+        [
+            ("list products", "availability"),
+            ("show me your product catalog", "availability"),
+            ("what CTV inventory do you have?", "availability"),
+            ("available impressions for display", "availability"),
+            ("how much does video inventory cost?", "pricing"),
+            ("what is the CPM for CTV?", "pricing"),
+            ("create deal for 5M impressions", "deal"),
+            ("I want to create a deal for 5M display impressions", "deal"),
+            ("book a campaign for next quarter", "deal"),
+            ("make a deal for sports inventory", "deal"),
+            ("that's too expensive, can you do $30?", "negotiation"),
+            ("how about $25 CPM instead?", "negotiation"),
+        ],
+    )
+    def test_prompt_routes_to_expected_intent(self, chat, prompt, expected_type):
+        result = chat.process_message(prompt)
+        assert result["type"] == expected_type, (
+            f"'{prompt}' routed to '{result['type']}', expected '{expected_type}'"
+        )
+
+    def test_general_handler_example_prompts_do_not_self_refer(self, chat):
+        """Example prompts printed by the general handler should not route to general."""
+        general_response = chat.process_message("hello")
+        assert general_response["type"] == "general"
+
+        example_lines = [
+            line.strip().lstrip("- ").strip('"')
+            for line in general_response["text"].splitlines()
+            if line.strip().startswith("- ")
+        ]
+        for example in example_lines:
+            result = chat.process_message(example)
+            assert result["type"] != "general", (
+                f"General handler's own example '{example}' routes back to general"
+            )


atc964 (approved):
Independently verified: clean 3-file diff, 13/13 new routing tests, full suite 1393/0 on the merge, and a live repro of the issue #46 script routes correctly ('list products' -> availability, deal phrasing -> deal, controls unchanged). Note for the record: this fixes the routing half of #46; the deploy.sh smoke-check gap stays open separately. One cosmetic nit, no action needed: the integration assertion matches '"type": "general"' with a literal space, which would pass vacuously under compact JSON serialization — the new unit tests are the real guard.

PR #48, opened 2026-08-10
fix: remove hardcoded US-scoped MEMORY_LLM_MODEL from deploy.sh

### LINKED ISSUE #45: deploy.sh hardcodes a US-scoped inference profile for MEMORY_LLM_MODEL instead of inheriting DEFAULT_LLM_MODEL
## Summary

`infra/aws/agentcore/deploy.sh` accepts `AWS_REGION`, but unconditionally injects a fixed, US-prefixed Bedrock inference profile as the memory model. This bypasses the parameterised model configuration the script uses everywhere else, and leaves no override path for the caller.

## Observed: the hardcoded value

Both runtime configurations, pinned at `2bb7ebce45891e3bdddd7a43198cb81d710b0916` (v2.4.1); also present on `main` at the same line numbers as of 7 August 2026:

```bash
--env "MEMORY_LLM_MODEL=bedrock/us.amazon.nova-lite-v1:0"
```

- [`deploy.sh#L483`](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/infra/aws/agentcore/deploy.sh#L483) — MCP runtime
- [`deploy.sh#L567`](https://github.com/IABTechLab/seller-agent/blob/2bb7ebce45891e3bdddd7a43198cb81d710b0916/infra/aws/agentcore/deploy.sh#L567) — HTTP runtime

The asymmetry with the surrounding configuration:

| Variable | Line | Overridable |
|---|---|---|
| `REGION` | L30 | yes — `${AWS_REGION:-us-west-2}` |
| `LLM_MODEL` → `DEFAULT_LLM_MODEL`, `MANAGER_LLM_MODEL` | L33 | yes — `${DEFAULT_LLM_MODEL:-...}` |
| `MEMORY_LLM_MODEL` | L483, L567 | **no** |

## Observed: the profile is rejected in `ap-southeast-2`

```
us.amazon.nova-lite-v1:0     us-west-2        -> OK
apac.amazon.nova-pro-v1:0    ap-southeast-2   -> OK
apac.amazon.nova-lite-v1:0   ap-southeast-2   -> OK
us.amazon.nova-lite-v1:0     ap-southeast-2   -> ValidationException:
                                                 The provided model identifier is invalid.
```

`ValidationException`, not `AccessDeniedException` — the identifier is not resolvable in that region, as distinct from being un-entitled.

`aws bedrock list-inference-profiles --region ap-southeast-2` returns `apac.amazon.nova-{micro,lite,pro}-v1:0` and `global.amazon.nova-2-lite-v1:0`. No `us.`-prefixed profiles are present.

The error text carries no indication of a region or prefix mismatch.

## Observed: the value reaches the deployed runtime

A default deployment does provision AgentCore Memory, so this is not a dormant setting.

<details>
<summary>Execution path (expand)</summary>

1. `agentcore configure` provisions AgentCore Memory unless `--disable-memory` is passed. `deploy.sh` does not pass it — the `configure_args` array is `-e`, `-n`, `-rf`, `-p`, `-r`, `--non-interactive`, `--deployment-type container`, plus VPC flags only under `STORAGE_TYPE=postgres`.
2. The toolkit injects `BEDROCK_AGENTCORE_MEMORY_ID` into the runtime environment at launch (`operations/runtime/launch.py:574`).
3. `settings.py:261` accepts `BEDROCK_AGENTCORE_MEMORY_ID` as valid embedder configuration, so the `CREW_MEMORY_ENABLED` guard does not disable memory.
4. `http_main.py:458` therefore applies `patches/crewai_agentcore_memory.py`, which reads `MEMORY_LLM_MODEL` in `Memory.__init__` (lines 120–123).

Verified against `bedrock-agentcore-starter-toolkit==0.3.4`, the version `deploy.sh` pins — note that it installs this only when `agentcore` is absent from `PATH` (L446, L529), so an environment with a different toolkit version may not share the same memory-provisioning default.

</details>

## Inferred

Not established, and flagged as such:

- **Scope beyond `ap-southeast-2`.** I have verified one non-US region. The `us.` prefix denotes US-scoped cross-region inference, so other non-US regions are likely affected, but I have not tested them.
- **Failure point.** `Memory.__init__` is wrapped in `try/except Exception` (`crewai_agentcore_memory.py:131`) which logs and continues, and LiteLLM generally resolves model strings lazily. The symptom may therefore be a startup failure, a failure at the first memory operation during crew execution, or a logged degradation. I have confirmed the invalid identifier is injected into the runtime configuration; I have not completed an end-to-end non-US deployment.

Either way, the configured memory model is not valid for the target region.

## Suggested remedy

**Preferred: remove the two lines.** The patch already falls back correctly:

```python
bedrock_model = os.environ.get(
    "MEMORY_LLM_MODEL",
    os.environ.get("DEFAULT_LLM_MODEL", "bedrock/us.amazon.nova-lite-v1:0"),
)
```

With `MEMORY_LLM_MODEL` unset, memory inherits `DEFAULT_LLM_MODEL` - which `deploy.sh` already parameterises via `${LLM_MODEL}`. Deleting L483 and L567 restores that inheritance with no other change.

**Alternative: make it caller-configurable.** `MEMORY_LLM_MODEL="${MEMORY_LLM_MODEL:-bedrock/us.amazon.nova-lite-v1:0}"` restores an override path, but retains a US-prefixed default that remains wrong for non-US callers who do not set it. The inheritance approach avoids that.

Where an explicit memory model is wanted, `apac.amazon.nova-lite-v1:0` is an available regional Nova Lite profile in `ap-southeast-2`.

One caution on a third option: `global.amazon.nova-2-lite-v1:0` is present in both regions and would resolve the immediate error, but `global.` profiles route to whichever region has capacity. For deployments with data-residency requirements that substitutes a harder problem for an easier one.

## Reproduction

The invalid model identifier can be reproduced without deploying seller-agent:

```bash
aws bedrock-runtime converse --region ap-southeast-2 \
  --model-id us.amazon.nova-lite-v1:0 \
  --messages '[{"role":"user","content":[{"text":"say ok"}]}]'
```

## Acceptance criteria

- A deployment in a region where no `us.`-prefixed profile is available (verified: `ap-southeast-2`) receives a memory model valid for that region, either by inheritance or by explicit override.
- The deployment script has no hardcoded region-scoped model identifier without an override path.

garvitkaushik-123:
## Summary

`deploy.sh` unconditionally injects `MEMORY_LLM_MODEL=bedrock/us.amazon.nova-lite-v1:0` into both the MCP and HTTP runtime configurations (lines 483 and 567). The `us.`-prefixed inference profile is not resolvable in non-US regions — verified as a `ValidationException` in `ap-southeast-2` — and there is no override path for the caller.

This is asymmetric with the surrounding configuration: `DEFAULT_LLM_MODEL` and `MANAGER_LLM_MODEL` are both parameterized via `${LLM_MODEL}`, but `MEMORY_LLM_MODEL` was hardcoded.

## Changes

Removed the two `--env "MEMORY_LLM_MODEL=..."` lines from `deploy_mcp_runtime()` and `deploy_http_runtime()`.

The memory patch (`patches/crewai_agentcore_memory.py:120-123`) already has the correct fallback chain:

```python
bedrock_model = os.environ.get(
    "MEMORY_LLM_MODEL",
    os.environ.get("DEFAULT_LLM_MODEL", "bedrock/us.amazon.nova-lite-v1:0"),
)
```

With `MEMORY_LLM_MODEL` unset, memory inherits `DEFAULT_LLM_MODEL` — which `deploy.sh` already parameterizes via `${LLM_MODEL}`. Non-US deployments receive a region-valid model without any additional change. Callers who want an explicit memory model can still set `MEMORY_LLM_MODEL` in their environment.

## Test plan

- [x] Verified no other references to `MEMORY_LLM_MODEL` in `deploy.sh` after the change
- [x] Confirmed the Python fallback chain in `crewai_agentcore_memory.py` inherits `DEFAULT_LLM_MODEL` when `MEMORY_LLM_MODEL` is unset
- [x] Full unit test suite passes (1393 tests, no regressions)

Closes #45

### DIFF
--- a/infra/aws/agentcore/deploy.sh
+++ b/infra/aws/agentcore/deploy.sh
@@ -480,7 +480,6 @@ deploy_mcp_runtime() {
     --env "ANTHROPIC_API_KEY=not-used-with-bedrock"
     --env "DATABASE_URL=sqlite:///:memory:"
     --env "CREW_MEMORY_ENABLED=true"
-    --env "MEMORY_LLM_MODEL=bedrock/us.amazon.nova-lite-v1:0"
   )
 
   if [[ "${INVENTORY_TYPE}" == "s3" ]]; then
@@ -564,7 +563,6 @@ deploy_http_runtime() {
     --env "ANTHROPIC_API_KEY=not-used-with-bedrock"
     --env "DATABASE_URL=sqlite:///:memory:"
     --env "CREW_MEMORY_ENABLED=true"
-    --env "MEMORY_LLM_MODEL=bedrock/us.amazon.nova-lite-v1:0"
   )
 
   if [[ "${INVENTORY_TYPE}" == "s3" ]]; then


atc964 (approved):
Verified: exactly the two hardcoded MEMORY_LLM_MODEL lines removed; the memory patch's fallback chain (MEMORY_LLM_MODEL -> DEFAULT_LLM_MODEL -> nova-lite) confirmed at patches/crewai_agentcore_memory.py:120-123, so non-US deployments now inherit the region-valid model and the env override still works. No residual references anywhere. Matches both acceptance criteria on #45.

jaanijuk:
Review note following the merge: removing the hardcoded `MEMORY_LLM_MODEL` resolves the non-US inference-profile problem cleanly, and the inheritance approach is the right shape.

One operational consequence worth recording: with no `MEMORY_LLM_MODEL` in the runtime environment, memory now inherits `DEFAULT_LLM_MODEL`, which `deploy.sh` sets from `${LLM_MODEL}` - Nova Pro on the default path. Memory analysis therefore moves from the former fixed Nova Lite model to the selected main agent model on every deployment, rather than only on non-US deployments. This is a model-selection and potentially cost-profile change, rather than a correctness concern.

The substitution applies only where `llm` is absent or is `"gpt-4o-mini"` (`crewai_agentcore_memory.py:118`); a different explicit `llm` value is not overwritten.

Relatedly, `crewai_agentcore_memory.py` retains `_read_only = True` with a comment that Nova Lite cannot handle `RememberSchema`. If memory now normally runs on the selected main model, that workaround may be worth reassessing separately, clearly outside this PR's scope.

I have an `ap-southeast-2` deployment planned and can confirm the regional fix end to end against `main`.

atc964:
Both points are right, thanks. On cost profile: agreed and worth stating plainly — with the hardcode gone, memory inherits DEFAULT_LLM_MODEL everywhere, which on the default path means the main model's pricing. We deliberately aren't re-pinning a cheap default: the old pin was a region-scoped inference profile, which is exactly what broke non-US deployments. Instead we'll document MEMORY_LLM_MODEL as the cost-tuning knob — set it explicitly in cost-sensitive deployments.

On _read_only: fair challenge — it existed because Nova Lite couldn't handle RememberSchema, and that constraint is gone when memory runs on the main model. We'll reassess. Your ap-southeast-2 end-to-end offer is gladly taken; results there would directly inform both decisions.

PR #49, opened 2026-08-11
fix: detect routing fallthrough in deploy.sh smoke check

garvitkaushik-123:
## Summary

Follow-up to #47, addressing the remaining smoke-check half of #46 as discussed in [the issue thread](https://github.com/IABTechLab/seller-agent/issues/46#issuecomment-5247189371).

The `_test_runtime` smoke check in `deploy.sh` only looks for error markers (`"error":`, `"exception":`, `Invocation failed`) and marks the test as passed when none are found. A chat-mode response that falls through to the general handler contains no error markers, so the smoke check cannot detect routing regressions — it silently passes on a misrouted response.

## Changes

- Add an optional fourth argument (`reject_type`) to `_test_runtime`. When set, the function checks whether the invoke output contains `"type": "<reject_type>"` and fails the test if it does.
- Pass `"general"` as `reject_type` for the two hardcoded `list products` chat-mode invocations (deploy modes `all` and `http`), so a future routing fallthrough is caught by the smoke check.
- Custom prompts (`--prompt` / `chat` deploy mode) are left without a reject type since the expected intent is unknown to the script.

## Test plan

- [x] `bash -n deploy.sh` — syntax valid
- [x] Verified the `reject_type` check is skipped when the argument is empty (crew mode calls, custom prompt calls)
- [x] Verified the check triggers on `"type": "general"` in output (would have caught the original #46 regression)

Closes the smoke-check follow-up from #46.

### DIFF
--- a/infra/aws/agentcore/deploy.sh
+++ b/infra/aws/agentcore/deploy.sh
@@ -738,6 +738,7 @@ if [[ "${DO_TEST}" == "true" ]]; then
     local agent_name="$1"
     local test_prompt="$2"
     local label="$3"
+    local reject_type="${4:-}"  # optional: fail if response contains this "type" value
     local max_retries=3
     local retry_wait=30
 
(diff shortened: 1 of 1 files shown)


atc964 (approved):
Verified: opt-in wiring correct (crew and custom-prompt paths unaffected), simulated detection catches the exact #46 failure shape, no new shellcheck findings, all checks green. Merging. Non-blocking follow-up idea: since python3 is already invoked in _test_runtime, parsing the response JSON instead of literal-grep would survive serialization drift — the equivalent assertion from #47 has the same trait, best updated together.

PR #50, opened 2026-08-11
fix: enforce idempotency on POST /api/v1/quotes, add conflict detection

### LINKED ISSUE #44: POST /api/v1/quotes requires idempotency_key but creates duplicate quotes on replay
`POST /api/v1/quotes` requires an `idempotency_key` (omitting it returns 422) but replaying the same key mints a new quote each time rather than returning the original. `POST /api/v1/deals` honours the same field correctly in the same build, which is what makes this look like an asymmetry rather than unfinished work.

**Environment.** seller-agent **v2.4.1** (commit `2bb7ebce`), primitives v0.5.0 (`5fdbe57`), WSL2 Ubuntu 24.04, Python 3.12.3. Also observed at **v2.3.3** and **v2.4.0** in a clean Ubuntu 24.04 container.

## Observed

The key is sent in the request body throughout.

**1. The field is required:**

```
POST /api/v1/quotes  {product_id, deal_type, impressions}   -> HTTP 422
```

**2. Four calls, one key, four quotes.** All four requests were byte-identical: same `idempotency_key`, `product_id`, `deal_type` and `impressions`, same anonymous buyer identity. Only the returned id changed.

```
call 1  HTTP 200  quote_id: qt-37f15263567f
call 2  HTTP 200  quote_id: qt-d9f805aff19d
call 3  HTTP 200  quote_id: qt-94a3a765b8ec
call 4  HTTP 200  quote_id: qt-935747361dfb
```

**3. Same key, different payload, accepted without error:**

```
{"idempotency_key":"<same as above>", "impressions":999999}  -> HTTP 200, qt-dfa11d3c6131
```

**Expected:** replaying an identical request with the same key returns the original quote and creates no second quote record.
**Actual:** each replay returns HTTP 200 with a new `quote_id`.

## Control: the same key behaves correctly on `/deals`

Run in the same process, minutes apart, against the same storage:

```
call 1  HTTP 200  deal_id: DEMO-E878783B822D
call 2  HTTP 200  deal_id: DEMO-E878783B822D
call 3  HTTP 200  deal_id: DEMO-E878783B822D
```

So the mechanism works in this build. It just isn't applied to quotes.

## In source

- `src/ad_seller/interfaces/api/routers/deals.py:94-115` documents **Idempotency (FD-12)** and implements it: it stores `idempotency:deal:{key}`, and on replay returns the existing deal "without minting a second one (no duplicate side effect)".
- `src/ad_seller/interfaces/api/routers/quotes.py` contains **no reference to idempotency at all**, though the request model requires the field.

## Why it's worth closing

A required `idempotency_key` is an invitation to build retry logic against it. A client that retries a timed-out quote request (the exact case the field exists for) silently creates duplicate quote records instead of returning the original quote.

Because price verification checks active quotes for the buyer and product (confirmed in #41), it would be useful to confirm whether duplicate quotes can affect verification behaviour. We have not tested that interaction.

## Questions rather than prescriptions

1. Is FD-12 intended to cover `/quotes` as well as `/deals`, with quote idempotency simply not yet implemented? If so, this is just a pointer to where the pattern is missing.

2. If quotes are deliberately excluded (because a quote is non-binding and cheap to re-mint) should the field be **optional** there rather than required? A mandatory field that is ignored is the part that misleads.

3. For a reused key with a different payload, should the API return a conflict (for example 409), treat it as a distinct request, or follow another documented rule?

Happy to retest whichever direction you take. The `deals.py` implementation appears structurally reusable for quotes, and I'm glad to raise a PR if that would help - although I'd rather not guess between questions 1 and 2 first.

garvitkaushik-123:
## Summary

Fixes #44. `POST /api/v1/quotes` requires `idempotency_key` (422 without it) but never honored it — replaying an identical request minted a new quote on every call. `POST /api/v1/deals` implements FD-12 correctly on the same surface, which is what made this look like an asymmetry rather than unfinished work.

Per [the maintainer's confirmation on the issue](https://github.com/IABTechLab/seller-agent/issues/44#issuecomment-5246886856):
1. A prior quote is a covered money-mutating operation under FD-12 — the gap is a bug, not a deliberate exclusion.
2. The key is intentionally required, no silent auto-minting.
3. A reused key with a different payload should return `idempotency_conflict`, mapped to HTTP 409.

The comment also flagged that `/deals` only detects *replay* (same key → same deal) but never detects *conflict* (same key, different body silently books whichever quote wins the race) — "welcome on both surfaces."

## Changes

**`contract_mappers.py`** — two new helpers:
- `idempotency_conflict_detail()`, mirroring the existing `unsupported_capability_detail()` pattern, builds the shared `ErrorDetail(error=ErrorCode.IDEMPOTENCY_CONFLICT)` envelope.
- `request_payload_hash()`, a stable sha256 over `request.model_dump(mode="json", exclude={"idempotency_key"})` — used to tell a true replay from a conflicting reuse.

**`routers/quotes.py`** — mirrors the `/deals` idempotency pattern under an `idempotency:quote:{key}` storage namespace:
- Replay with a matching payload hash returns the original quote (via `quote_service.get_quote`), no second quote minted.
- Replay with a different hash raises `409 idempotency_conflict`.

**`routers/deals.py`** — closes the conflict-detection gap: previously any two different bodies under the same key would just book whichever request completed first booking (silent, no error). Now a payload-hash mismatch under a reused key returns `409 idempotency_conflict`, same as quotes.

## Test plan

- [x] 3 new tests in `test_quote_endpoints.py::TestQuoteIdempotency` — replay returns same quote (and only one quote record persisted), reused key + different body → 409, different keys → separate quotes
- [x] 1 new test in `test_deal_booking_endpoints.py::TestBookDeal` — reused key against a different `quote_id` → 409
- [x] Existing replay/error-path tests on both endpoints pass unchanged
- [x] Full unit suite: 1397 passed, no regressions
- [x] `ruff check` / `ruff format --check` clean

Closes #44

### DIFF
--- a/src/ad_seller/interfaces/api/contract_mappers.py
+++ b/src/ad_seller/interfaces/api/contract_mappers.py
@@ -16,6 +16,8 @@ dollars; the shared contract prices in :class:`Money` (integer micros).
 Conversion happens here and nowhere else.
 """
 
+import hashlib
+import json
 from datetime import date, datetime
 from typing import Any, Optional
 
@@ -109,6 +111,30 @@ def unsupported_capability_detail(
         unsupported=capabilities,
     )
     return detail.model_dump(mode="json")
+
+
+def idempotency_conflict_detail(message: str = "") -> dict[str, Any]:
+    """Build the ``{"error": "idempotency_conflict", ...}`` inner detail (FD-12).
+
+    Returned as the ``detail`` of a FastAPI ``HTTPException(status_code=409)``
+    when an ``idempotency_key`` is reused with a different request body.
+    """
+    detail = ErrorDetail(
+        error=ErrorCode.IDEMPOTENCY_CONFLICT,
+        message=message or "idempotency_key was reused with a different request body.",
+    )
+    return detail.model_dump(mode="json")
+
+
+def request_payload_hash(payload: dict[str, Any]) -> str:
+    """Stable hash of a request payload, for idempotency-key conflict detection.
+
+    Callers pass ``request.model_dump(mode="json", exclude={"idempotency_key"})``
+    so the key itself never affects its own conflict check. ``sort_keys``
+    makes the hash independent of dict/field ordering.
+    """
+    encoded = json.dumps(payload, sort_keys=True, default=str).encode()
+    return hashlib.sha256(encoded).hexdigest()
 
 
 # ---------------------------------------------------------------------------
--- a/src/ad_seller/interfaces/api/contract_mappers.py
+++ b/src/ad_seller/interfaces/api/contract_mappers.py
@@ -137,6 +137,30 @@ def request_payload_hash(payload: dict[str, Any]) -> str:
     return hashlib.sha256(encoded).hexdigest()
 
 
+def idempotency_buyer_scope(identity: Any) -> str:
+    """Stable per-buyer scope string for idempotency-key namespacing (FD-12).
+
+    FD-12 scopes an ``idempotency_key`` to a buyer/seller pair — a global
+    key namespace lets two different buyers collide on the same
+    client-chosen key. Mirrors the specificity order of
+    :meth:`BuyerContext.get_pricing_key` so a request-body identity
+    (``buyer_identity`` / an API key's ``.identity``) resolves to the same
+    scope wherever it is used.
+
+    ``identity`` is any object exposing optional ``advertiser_id`` /
+    ``agency_id`` / ``seat_id`` attributes (or ``None``).
+    """
+    if identity is None:
+        return "public"
+    if getattr(identity, "advertiser_id", None):
+        return f"advertiser:{identity.advertiser_id}"
+    if getattr(identity, "agency_id", None):
+        return f"agency:{identity.agency_id}"
+    if getattr(identity, "seat_id", None):
+        return f"seat:{identity.seat_id}"
+    return "public"
+
+
 # ---------------------------------------------------------------------------
 # Quote surface
 # ---------------------------------------------------------------------------
--- a/src/ad_seller/interfaces/api/routers/deals.py
+++ b/src/ad_seller/interfaces/api/routers/deals.py
@@ -95,7 +95,10 @@ async def book_deal(
     ``idempotency_key``. A replay with a key already used for an identical
     body returns the same Deal without minting a second one (no duplicate
     side effect); the same key reused with a different body is an
-    ``idempotency_conflict`` (HTTP 409).
+    ``idempotency_conflict`` (HTTP 409). The key is scoped per buyer
+    (FD-12) — API key identity takes priority over the self-asserted
+    ``buyer_identity`` body field, matching the priority order used to
+    resolve buyer identity elsewhere.
     """
     from ....storage.factory import get_storage
 
@@ -105,7 +108,11 @@ async def book_deal(
     # response, no duplicate booking (FD-12). Kept at the wire edge so
     # deal_service stays untouched. Defensive against storage backends/
     # mocks that do not implement the generic get/set KV methods.
-    idem_storage_key = f"idempotency:deal:{request.idempotency_key}"
+    buyer_identity = (
+        api_key_record.identity if api_key_record is not None else request.buyer_identity
+    )
+    buyer_scope = cm.idempotency_buyer_scope(buyer_identity)
+    idem_storage_key = f"idempotency:deal:{buyer_scope}:{request.idempotency_key}"
     payload_hash = cm.request_payload_hash(
         request.model_dump(mode="json", exclude={"idempotency_key"})
     )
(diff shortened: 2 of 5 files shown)


atc964 (changes requested):
Thanks — the core of this is right: canonical payload hashing (model_dump + sort_keys, immune to key order and unknown fields), the 409 envelope matches the contract's idempotency_conflict vector, TTL and status codes are consistent with /deals, and covering both surfaces was exactly what #44 asked. Full suite green here (1397 passed). Two changes needed before merge:

1. Deals legacy-record migration (regression, reproduced locally): existing idempotency:deal:{key} records store a plain string deal_id, and the new code replays only when the record is a dict — so a retry that straddles this deploy falls through and books a duplicate deal, overwriting the record. Please handle the legacy shape: elif isinstance(prior, str) and prior: replay it without the conflict check (hash unknown for legacy records), or migrate on read.

2. Buyer scoping (contract requirement): FD-12 scopes keys per buyer/seller pair, but idempotency:quote:{key} is global — two buyers reusing the same key collide, and because the replay check runs before _verified_buyer_context, the second buyer's identity/tier verification is bypassed on the replay path. Please fold the verified buyer principal into the namespace (or the payload hash), and run buyer verification before the replay lookup. /deals has the same pre-existing gap — same pass or a follow-up, your call.

Noted, non-blocking: the get-then-set race (two concurrent same-key requests both create) is pre-existing on /deals and fine to leave for a follow-up issue with an atomic set-if-absent; fail-open on storage errors matches the deliberate /deals behavior; conflict-check-before-validation ordering is a defensible choice.

Happy to re-review quickly once those two land.

garvitkaushik-123:
Both addressed in ce7e9d9:

**1. Legacy record migration.** Added an `elif isinstance(prior, str) and prior:` branch in `deals.py` that replays a pre-migration string-shaped record as-is — no hash to compare against, so no conflict check on it (matches the trust level that record was written under). Confirmed with `test_legacy_string_record_replays_without_conflict`: a bare-string record seeded at the old shape replays cleanly with no duplicate deal minted.

**2. Buyer scoping.**
- `quotes.py`: moved `_verified_buyer_context` before the idempotency lookup, and scoped the storage key with the resulting `BuyerContext.get_pricing_key()` (advertiser > agency > seat > public). A replay now can't skip a different buyer's own verification, and the namespace is per-buyer.
- `deals.py`: `book_deal` has no registry-based tier verification to move — that's the separate, larger pre-existing gap you flagged as "your call." I scoped the key by resolved identity instead (API key identity when present, else the self-asserted `buyer_identity` body field — same priority order used elsewhere), which stops the cross-buyer collision without pulling full trust verification into this PR. Happy to open a follow-up for that if you want it tracked separately.
- Added `idempotency_buyer_scope()` to `contract_mappers.py`, mirroring `get_pricing_key()`'s specificity order so both surfaces land on the same scope for the same buyer.

Tests: 3 new (quote buyer-scope isolation, deal buyer-scope isolation, legacy-record replay). Full suite: 1400 passed, no regressions. `ruff check` / `format --check` clean.

atc964 (changes requested):
Fast turnaround appreciated — the buyer-scoping fix is verified working: identity resolution now precedes the lookup, keys scope via get_pricing_key(), and my two-buyers-same-key adversarial check shows no cross-tenant replay on either surface. That half is done.

One blocker remains, and it's subtle: the legacy migration branch is correct in shape but unreachable, because the scoping change altered the lookup key out from under it. Current main writes legacy records at idempotency:deal:{key} (bare, unscoped); this head only ever reads idempotency:deal:{scope}:{key}, and scope is never empty. Reproduced on ce7e9d9: seeded idempotency:deal:idem-legacy exactly as main writes it, replayed, and got a fresh duplicate deal — the same regression as before, relocated. The new test passes only because it seeds the record at the post-PR key shape (idempotency:deal:public:idem-legacy), which production never wrote.

Tightly scoped ask: (1) in book_deal, on a scoped-key miss, add a fallback read of the legacy unscoped idempotency:deal:{key} and route it through your str-replay branch (no conflict check — no stored hash). Since legacy TTL is 24h, that fallback is removable one TTL window after deploy. (2) Fix the legacy test to seed idempotency:deal:idem-legacy with no scope segment so it reproduces the true production shape.

Everything else is clean: merge-clean over #49, full suite 1400 passing here, diff hygiene good. This is the last item — happy to re-review same-day again.

garvitkaushik-123:
Good catch — confirmed the repro and fixed in 25373d5.

Added a fallback in `book_deal`: on a scoped-key miss, read the legacy unscoped `idempotency:deal:{key}` before concluding there's no prior booking. A hit there routes through the existing string-replay branch unchanged — no stored hash on legacy records, so no conflict check, same trust level they were written under. Noted it's removable one TTL window (24h) after this deploys.

Fixed `test_legacy_string_record_replays_without_conflict` to seed `idempotency:deal:idem-legacy` (no scope segment) — the actual shape `main` writes — instead of the post-PR scoped shape that let the regression through review. Reran it against the fix: replays cleanly, no duplicate deal.

Full suite: 1400 passed. `ruff check` / `format --check` clean.

atc964 (approved):
Verified fixed by reproduction: the legacy unscoped record now replays through the fallback (seeded exactly as main writes it — replay returned the stored deal, booking service never invoked, store unchanged), the fix is load-bearing (reverting the hunk fails your corrected legacy test), scoped precedence is right, and the buyer-scoping half is byte-identical to the round-2 verified version. Full suite 1400/0 here, CI green. The 24h global-replay window for legacy records is inherent and time-boxed — exactly the intended design. Thanks for three fast, honest iterations — merging.

PR #52, opened 2026-08-11
fix: enforce idempotency on POST /api/v1/negotiations/messages

### LINKED ISSUE #51: POST /api/v1/negotiations/messages requires idempotency_key but never honors it — retry burns a round from the negotiation's max_rounds budget
## Summary

`POST /api/v1/negotiations/messages` accepts the shared `NegotiationMessage` contract type, which requires `idempotency_key` (inherited from `IdempotentRequest`, same base class as `QuoteRequest`/`DealBookingRequest` — omitting it is a 422). Neither the router (`interfaces/api/routers/negotiation.py`) nor `negotiation_service.counter_proposal` / `apply_terminal_action` ever read the field. This is the same bug class as #44 (closed via #50), on a different endpoint that wasn't covered by that fix.

Unlike quotes, where a duplicate is just an extra ephemeral record, a negotiation has a hard `max_rounds` budget (3–6 depending on tier — see `models/negotiation.py`). Each `counter_proposal` call unconditionally appends a new round and increments `round_number`, with no replay check. A client retry of a timed-out counter — the exact case `idempotency_key` exists to make safe — burns an extra round from that budget instead of returning the original response. Enough retries can push a legitimate negotiation past `max_rounds` into an unintended REJECT/walk-away (`engines/negotiation_engine.py:181-190`).

## Reproduction

No server needed — via the FastAPI app directly, same harness pattern as `tests/unit/test_negotiation_lowball_counter.py` (mocked storage, real `negotiation_service`):

```python
body = {
    "idempotency_key": "idem-same-key-retried",
    "action": "counter",
    "proposal_id": "prop-repro",
    "buyer_price": {"amount_micros": 30_000_000, "currency": "USD"},
}

first = await client.post("/api/v1/negotiations/messages", json=body)
second = await client.post("/api/v1/negotiations/messages", json=body)  # identical key + body
```

Output:

```
first status: 200   round_number: 1   rounds_remaining: 2
second status: 200  round_number: 2   rounds_remaining: 1
rounds persisted in storage: 2
```

Two identical requests (same key, same body) minted two rounds and consumed two units of the round budget. `negotiation:prop-repro` in storage has 2 entries in `rounds` from what should have been one buyer move.

## In source

- `NegotiationMessage` requires `idempotency_key`: [`iab_agentic_primitives/protocol/negotiation.py:76-77`](https://github.com/IABTechLab/iab-agentic-primitives) (`IdempotentRequest` base, `_base.py`) — "a retried counter must not consume an [extra round]" per the base class's own docstring.
- `POST /api/v1/negotiations/messages` handler: [`src/ad_seller/interfaces/api/routers/negotiation.py:93-158`](https://github.com/IABTechLab/seller-agent/blob/dce0f5251774aae35fa17f7d84806c2ca5f4e2ab/src/ad_seller/interfaces/api/routers/negotiation.py#L93-L158) — `message.idempotency_key` is never referenced.
- `negotiation_service.counter_proposal`: [`src/ad_seller/services/negotiation_service.py:189-`](https://github.com/IABTechLab/seller-agent/blob/dce0f5251774aae35fa17f7d84806c2ca5f4e2ab/src/ad_seller/services/negotiation_service.py#L189) — appends a round unconditionally on every call, no idempotency check anywhere in the file (`grep idempotency` = 0 hits).
- `max_rounds` enforcement: [`src/ad_seller/engines/negotiation_engine.py:181-190`](https://github.com/IABTechLab/seller-agent/blob/dce0f5251774aae35fa17f7d84806c2ca5f4e2ab/src/ad_seller/engines/negotiation_engine.py#L181-L190) — exceeding it triggers a REJECT.
- Existing tests (`tests/unit/test_negotiation_message_endpoint.py`) pass `idempotency_key` in every request body (required to pass validation) but mock `negotiation_service.counter_proposal` directly, so none of them exercise replay behavior against real storage — the gap has no test coverage in either direction.

## Scope check: other endpoints

- `POST /api/v1/quotes` and `POST /api/v1/deals` — fixed in #50.
- `POST /api/v1/change-requests` — uses the seller's own internal `CreateChangeRequestModel` (`interfaces/api/schemas.py:355`), not the shared contract's `ChangeRequestBody`. No `idempotency_key` field on this endpoint at all, so no false-advertised guarantee here — not affected.
- Legacy `POST /proposals/{proposal_id}/counter` (`CounterOfferRequest`) — also a seller-internal model with no `idempotency_key` field. Not affected.
- `POST /api/v1/negotiations/messages` is the only other shared-contract, money-mutating, `idempotency_key`-required surface, and it's the one still open.

## Suggested remedy

Mirror the pattern from #50: on `action in PRICED_ACTIONS` (counter/final_offer — the price-moving paths), check a `idempotency:negotiation:{scope}:{key}` storage record before calling `counter_proposal`; on a hash match, replay the last round's response instead of appending a new one; on a hash mismatch, `idempotency_conflict` (409). `accept`/`reject` (`apply_terminal_action`) are lower risk since they're terminal and largely no-op on an already-terminal negotiation, but the same required field applies to those messages too and is equally unread.

Happy to raise a PR for this if it's confirmed as in-scope — the shape should closely follow #50's `contract_mappers.request_payload_hash` / `idempotency_conflict_detail` helpers, which already exist and are reusable here.

## Environment

seller-agent `main` @ `dce0f52` (post-#49/#50 merge), 2026-08-11.

garvitkaushik-123:
## Summary

Fixes #51. `NegotiationMessage` requires `idempotency_key` (inherited from `IdempotentRequest` — the same base class as `QuoteRequest`/`DealBookingRequest`, a 422 without it), but neither the router nor `negotiation_service` ever read the field. Same bug class as #44 (fixed for quotes/deals in #50), on a surface that wasn't covered by that fix.

This one is worse than the quotes case: negotiations have a hard `max_rounds` budget (3–6 depending on tier). Every `counter_proposal` call unconditionally appends a round with no replay check, so a client retry of a timed-out counter — the exact scenario `idempotency_key` exists to make safe — burns an extra round instead of returning the original response. Enough retries can push a legitimate negotiation past `max_rounds` into an unintended REJECT/walk-away.

Reproduced live before fixing (see #51 for the full repro): identical `idempotency_key` + identical body, two requests → second call minted `round_number=2` and `rounds_remaining` dropped 2→1 from a single retry.

## Changes

Mirrors the #50 pattern in `routers/negotiation.py`'s `post_negotiation_message`:

- Checks an `idempotency:negotiation:{buyer_scope}:{key}` storage record before running either branch — the field is required on every action (accept/reject/counter/final_offer), not just the priced ones, so the check runs once up front rather than duplicated per-branch.
- On a payload-hash match: replays the stored response instead of re-running the engine or re-applying a terminal move.
- On a payload-hash mismatch: `409 idempotency_conflict`, reusing `contract_mappers.idempotency_conflict_detail()` / `request_payload_hash()` from #50.
- Scoped per buyer via `BuyerContext.get_pricing_key()` (same specificity order as quotes/deals) — a global namespace would let two buyers collide on the same client-chosen key.

## Test plan

- 3 new tests in `test_negotiation_message_endpoint.py::TestNegotiationMessageIdempotency`:
  - replay (same key + body) returns the identical response and the negotiation engine is called exactly once
  - reused key + different `buyer_price` → 409, engine still called only once
  - a record stored under a different buyer's scope doesn't short-circuit this buyer's request
- Two pre-existing tests in the same file didn't mock storage — harmless before this change (nothing in the request path touched it), but now that the endpoint has real storage-backed idempotency, an un-mocked call falls through to whichever storage backend is actually configured. Fixed both to mock storage explicitly, matching the convention already used in `test_deal_booking_endpoints.py` / `test_quote_endpoints.py`.
- Full suite: 1403 passed (1400 + 3 new). `ruff check` / `format --check` clean.

Closes #51

### DIFF
--- a/src/ad_seller/interfaces/api/routers/negotiation.py
+++ b/src/ad_seller/interfaces/api/routers/negotiation.py
@@ -103,6 +103,13 @@ async def post_negotiation_message(
     bug). Priced actions (counter/final_offer) run the seller's untouched
     negotiation engine; accept/reject are recorded as terminal rounds off
     the existing history.
+
+    **Idempotency (FD-12):** the request carries a required
+    ``idempotency_key``. A replay with a key already used for an identical
+    body returns the same round without consuming another one from the
+    negotiation's ``max_rounds`` budget; the same key reused with a
+    different body is an ``idempotency_conflict`` (HTTP 409). The key is
+    scoped per buyer, same as ``/api/v1/quotes`` and ``/api/v1/deals``.
     """
     # EP-5.2: cap the claimed tier at a verified ceiling. The shared
     # NegotiationMessage carries no agent_url (lib gap — QuoteRequest has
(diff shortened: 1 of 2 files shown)


atc964 (approved):
Verified: FD-12 coverage up-front for all four negotiation actions, canonical hashing (excludes the key, sort_keys), identity resolution before lookup with get_pricing_key scoping — the #50 lessons are all applied, and the namespace is brand-new so no legacy-record concerns. Load-bearing check confirmed: with main's router restored, your replay and conflict tests fail — they're real guards. Full suite 1403/0 here, CI green. Storing the full round response instead of an id-to-refetch is the right call for non-refetchable rounds. Merging. One follow-up we'll file separately: the legacy /proposals/{id}/counter route still burns rounds with no idempotency — pre-existing and outside your scope.

PR #54, opened 2026-08-11
fix: don't stack a higher-priority discount rule onto a price override

### LINKED ISSUE #53: PricingRulesEngine: a higher-priority discount rule's discount silently stacks onto a lower-priority rule's base_price_override
## Summary

`PricingRulesEngine.calculate_price` applies a matching rule's `base_price_override` as a flat replacement price, and the code comment states this is deliberate: `# Price override takes precedence`. In practice, if a *different*, higher-priority rule with only a `discount_percentage` (no override) matches the same request, that rule's discount is silently multiplied into the override price — the override does not take precedence, it gets discounted on top.

This directly affects revenue: a seller-configured flat negotiated rate ends up cheaper than configured whenever a broader tier/agency discount rule also matches and outranks it in priority.

## In source

[`src/ad_seller/engines/pricing_rules_engine.py:109-123`](https://github.com/IABTechLab/seller-agent/blob/47beb9b4e07794d51e1e08df017f1eb67e6f335c/src/ad_seller/engines/pricing_rules_engine.py#L109-L123):

```python
rule_discount = 0.0
for rule in matching_rules:
    if rule.base_price_override is not None:
        price = rule.base_price_override
        applied_rules.append(
            f"Rule '{rule.rule_name}': Price override ${rule.base_price_override}"
        )
        break  # Price override takes precedence

    if rule.discount_percentage > 0:
        rule_discount = max(rule_discount, rule.discount_percentage)

if rule_discount > 0:
    price = price * (1 - rule_discount)
    applied_rules.append(f"Rule discount: -{rule_discount * 100:.0f}%")
```

`matching_rules` is sorted by priority, descending ([`models/pricing_tiers.py:214-237`](https://github.com/IABTechLab/seller-agent/blob/47beb9b4e07794d51e1e08df017f1eb67e6f335c/src/ad_seller/models/pricing_tiers.py#L214-L237)), not by whether a rule has an override. Nothing in `PricingRule` ([`models/pricing_tiers.py:37-68`](https://github.com/IABTechLab/seller-agent/blob/47beb9b4e07794d51e1e08df017f1eb67e6f335c/src/ad_seller/models/pricing_tiers.py#L37-L68)) requires override rules to carry the highest priority.

The loop accumulates `rule_discount` from every discount-only rule visited *before* the break — including rules with a **higher priority than the override rule**, which are visited first precisely because they're higher priority. Once the override rule is reached, `price` is overwritten, the loop exits, but `rule_discount` is not reset, so the trailing `if rule_discount > 0` block re-applies it to the override price.

## Reproduction

```python
from ad_seller.engines.pricing_rules_engine import PricingRulesEngine
from ad_seller.models.pricing_tiers import TieredPricingConfig, PricingRule
from ad_seller.models.buyer_identity import BuyerContext, BuyerIdentity, AccessTier

config = TieredPricingConfig(seller_organization_id="test-seller")
config.rules = [
    PricingRule(
        rule_id="r-discount", rule_name="Agency-wide 20pct discount",
        priority=100, access_tier=AccessTier.AGENCY, discount_percentage=0.20,
    ),
    PricingRule(
        rule_id="r-override", rule_name="Special negotiated flat rate $50",
        priority=50, access_tier=AccessTier.AGENCY, base_price_override=50.0,
    ),
]

engine = PricingRulesEngine(config)
buyer = BuyerContext(identity=BuyerIdentity(agency_id="a1"), is_authenticated=True)
decision = engine.calculate_price(product_id="p1", base_price=100.0, buyer_context=buyer, volume=0)
print(decision.final_price)     # 40.0
print(decision.applied_rules)
```

Output:

```
final_price: 40.0
applied_rules: ['Tier discount: -10%', "Rule 'Special negotiated flat rate $50': Price override $50.0", 'Rule discount: -20%']
```

The seller configured a $50 flat rate for this agency. The buyer is quoted $40 — a further 20% off a price the seller intended to be final — because the agency-wide discount rule outranks the override rule in priority and gets evaluated (and its discount recorded) before the override is applied.

## Test coverage

`base_price_override` has zero references in `tests/unit/test_pricing_engine.py` — this interaction (an override rule co-existing with a higher-priority discount-only rule) has no test coverage in either direction.

## Suggested remedy

Once an override is applied, the accumulated `rule_discount` from rules visited earlier in the loop should not be applied to it — either reset `rule_discount = 0.0` at the point of the override (and skip the trailing discount block for that path), or restructure to check for an override across all matching rules *first*, independent of iteration order, before falling back to accumulating discounts. The comment's stated intent ("Price override takes precedence") should hold regardless of relative rule priority ordering.

Happy to raise a PR for this — the fix is a small, localized change to `calculate_price`.

## Environment

garvitkaushik-123:
## Summary

Fixes #53. `PricingRulesEngine.calculate_price` scans `matching_rules` (sorted by priority, descending) and breaks as soon as it finds a rule with `base_price_override`, with a comment saying the override "takes precedence." It does — for the running `price` variable. But a `discount_percentage` from a rule visited *earlier* in the same loop (a higher-priority rule with no override of its own) stays in the shared `rule_discount` accumulator and gets multiplied into the override price after the loop exits.

Verified against `docs/guides/pricing-rules.md` before filing #53 — it draws override and discount as mutually exclusive branches off the same decision point, and states an override "stops further rule evaluation." The current code does the opposite whenever a higher-priority discount-only rule happens to be scanned first.

## Concrete impact

A seller configures a $50 flat negotiated rate for an agency (priority 50), alongside an unrelated agency-wide 20% discount rule (priority 100, no override). The discount rule is evaluated first — it outranks the override rule — leaving `0.20` in `rule_discount`. The override rule is then reached, sets `price = 50`, and breaks. The stale `0.20` still gets applied afterward: the buyer lands at **$40** instead of the **$50** the seller configured. Revenue leak, silent.

## Fix

Track whether an override fired (`override_applied`) and skip the trailing discount-application block when it did. The discount-only path (no override in play, multiple discount rules competing) is untouched — still takes the highest discount, not stacked, as documented.

```python
rule_discount = 0.0
override_applied = False
for rule in matching_rules:
    if rule.base_price_override is not None:
        price = rule.base_price_override
        ...
        override_applied = True
        break

    if rule.discount_percentage > 0:
        rule_discount = max(rule_discount, rule.discount_percentage)

if rule_discount > 0 and not override_applied:
    price = price * (1 - rule_discount)
    ...
```

## Test plan

`base_price_override` had **zero** test coverage before this — that's how the bug survived since the initial commit. Added `TestPriceOverrideRules` in `test_pricing_engine.py`:

- override alone replaces the base price
- override wins even when outranked (lower priority) by a discount-only rule — the case that was broken
- override wins when it outranks the discount rule — the case that already worked, kept as a regression guard
- discount-only rules (no override) still apply the highest one, not stacked — confirms the fix didn't touch the unrelated path

Full suite: 1404 passed, no regressions. `ruff check` / `format --check` clean.

Closes #53

### DIFF
--- a/src/ad_seller/engines/pricing_rules_engine.py
+++ b/src/ad_seller/engines/pricing_rules_engine.py
@@ -123,8 +123,14 @@ class PricingRulesEngine:
         # A higher-priority rule's discount can already be sitting in
         # rule_discount by the time we hit the override below it in the
         # list — don't let it sneak back in and knock more off an override
-        # price that's supposed to be the final word.
-        if rule_discount > 0 and not override_applied:
+        # price that's supposed to be the final word. Reset it outright
+        # rather than just skipping the price multiply: rule_discount also
+        # feeds the rationale string below, and a leftover value there
+        # would have the rationale claim a discount that was never
+        # actually applied to the override price.
+        if override_applied:
+            rule_discount = 0.0
+        elif rule_discount > 0:
             price = price * (1 - rule_discount)
             applied_rules.append(f"Rule discount: -{rule_discount * 100:.0f}%")
 
--- a/src/ad_seller/engines/pricing_rules_engine.py
+++ b/src/ad_seller/engines/pricing_rules_engine.py
@@ -107,18 +107,24 @@ class PricingRulesEngine:
         )
 
         rule_discount = 0.0
+        override_applied = False
         for rule in matching_rules:
             if rule.base_price_override is not None:
                 price = rule.base_price_override
                 applied_rules.append(
                     f"Rule '{rule.rule_name}': Price override ${rule.base_price_override}"
                 )
-                break  # Price override takes precedence
+                override_applied = True
+                break  # Price override takes precedence, stop evaluating rules
 
             if rule.discount_percentage > 0:
                 rule_discount = max(rule_discount, rule.discount_percentage)
 
-        if rule_discount > 0:
+        # A higher-priority rule's discount can already be sitting in
+        # rule_discount by the time we hit the override below it in the
+        # list — don't let it sneak back in and knock more off an override
+        # price that's supposed to be the final word.
+        if rule_discount > 0 and not override_applied:
             price = price * (1 - rule_discount)
             applied_rules.append(f"Rule discount: -{rule_discount * 100:.0f}%")
 
--- a/tests/unit/test_pricing_engine.py
+++ b/tests/unit/test_pricing_engine.py
@@ -13,9 +13,9 @@ Focuses on:
 import pytest
 
 from ad_seller.engines.pricing_rules_engine import PricingRulesEngine
-from ad_seller.models.buyer_identity import BuyerContext, BuyerIdentity
+from ad_seller.models.buyer_identity import AccessTier, BuyerContext, BuyerIdentity
 from ad_seller.models.core import PricingModel
-from ad_seller.models.pricing_tiers import TieredPricingConfig
+from ad_seller.models.pricing_tiers import PricingRule, TieredPricingConfig
 
 
 @pytest.fixture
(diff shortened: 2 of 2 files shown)


atc964 (changes requested):
Reproduced the bug on main exactly as the issue describes — override configured at $50, discount rule leaks in, buyer pays $40. Silent revenue leak, docs unambiguously say override-wins, and the fix's override_applied gate is the right minimal shape. Adjacent cases verified here: highest-priority override wins among two, volume discount still applies post-override per the documented flow, floor still catches zero/negative. Full suite 1407/0 on the merge, and your override test is genuinely load-bearing.

One change needed before merge, and it's a money-path provenance issue: the fix suppresses the discount from applied_rules but not from the rationale. rule_discount still holds the stale value when _build_rationale runs, so the corrected path emits 'Custom rule: -20% ... Final price: $50.00' — asserting a discount that was not applied. That string flows into quotes, negotiated-deal text, and the MCP/API surfaces, where a negotiating counterparty could anchor on it. Exact ask: zero out rule_discount in the override branch (or gate the rationale segment on not override_applied), plus one test asserting 'Custom rule' is absent from the rationale on the override path.

Noted, not blocking: the tier-discount segment lingers in rationale the same way on main (pre-existing), and base_price_override accepts negatives at the model layer with the floor catching it downstream — both worth follow-ups. Happy to re-review same-day.

garvitkaushik-123:
Good catch — fixed in 3593652.

rule_discount is now reset to 0.0 outright when override_applied is true, instead of just gating the price-multiply block. Same variable feeds both applied_rules and the rationale string, so one reset covers both — the rationale on the override path no longer mentions "Custom rule" at all now.

Added test_rationale_does_not_claim_an_unapplied_discount asserting exactly that: "Custom rule" absent from the rationale, "Final price: \$50.00" present, on the same higher-priority-discount-vs-lower-priority-override setup from the original bug.

Full suite: 1405 passed. `ruff check` / `format --check` clean.

garvitkaushik-123:
Updated as per the requested changes above.

atc964 (approved):
Re-verified at 3593652: the reset sits ahead of both consumers (applied_rules via the mutually-exclusive elif, rationale via the rule_discount gate), the round-1 repro now yields $50 with a clean rationale, and the discount-only path still reports its discounts — no over-suppression. Your new test is genuinely load-bearing: with the reset reverted it fails on exactly the phantom-rationale leak while the other pricing tests pass. Suite 1439/0 on current main here, CI green. The lingering tier-segment and savings-line rationale quirks on the override path are pre-existing and tracked our side. Merging — thanks for the quick, precise turnaround.

PR #56, opened 2026-08-11
fix: reject migrating an already-migrated deal, don't fork lineage

### LINKED ISSUE #55: migrate_deal has no guard against migrating an already-migrated/deprecated deal — silently orphans the prior replacement and forks lineage
## Summary

`POST /api/v1/deals/{deal_id}/migrate` has no check on the target deal's current status before mutating it. Calling it twice on the same `deal_id` — a duplicate client call, a retry, or an operator error — silently orphans the first replacement deal (still fully active in storage, no longer reachable via lineage) and rewrites the original deal's `replacement_deal_id` to point at a second, unrelated replacement.

This is inconsistent with the sibling endpoint: `deprecate_deal` (`services/deal_service.py:1440-1449`) explicitly rejects acting on an already-deprecated deal with `409 already_deprecated`. `migrate_deal` performs a strictly more state-mutating operation (creates a new deal *and* deprecates the old one) with no equivalent guard.

Per the MCP tools reference (`docs/reference/mcp-tools.md:47`): *"Replace an existing deal with a new one. **Creates lineage tracking.**"* — the whole point of this endpoint is a trackable chain. A double-migrate breaks that chain instead.

## Reproduction

```python
store = {}
storage = AsyncMock()
storage.get_deal = AsyncMock(side_effect=lambda did: store.get(did))
storage.set_deal = AsyncMock(side_effect=lambda did, data: store.__setitem__(did, data))

store["DEAL-ORIG"] = {
    "deal_id": "DEAL-ORIG", "deal_type": "PD", "status": "confirmed",
    "product_id": "p1", "actual_price_cpm": 30.0, "impressions": 1_000_000,
}

with patch("ad_seller.storage.factory.get_storage", return_value=storage):
    result1 = await deal_service.migrate_deal("DEAL-ORIG", req1)   # first migration
    first_replacement = result1["new_deal_id"]

    result2 = await deal_service.migrate_deal("DEAL-ORIG", req2)   # migrated AGAIN
    second_replacement = result2["new_deal_id"]
```

Output:

```
First migration -> replacement: DEMO-5977777A7719
DEAL-ORIG status after 1st migrate: deprecated
Second migration -> replacement: DEMO-7B94A5F00180
DEAL-ORIG replacement_deal_id now points to: DEMO-7B94A5F00180

First replacement deal still in storage: True
Is first replacement referenced by ANYTHING now? False

Forward lineage from DEAL-ORIG: ['DEMO-7B94A5F00180']
First replacement DEMO-5977777A7719 reachable via lineage walk? False
```

`DEMO-5977777A7719` — the first replacement — is still a live, `confirmed`-status deal record in storage. It has no incoming or outgoing lineage pointer from anything: `get_deal_lineage("DEAL-ORIG")` only ever finds the second replacement. Any buyer-side or seller-side tooling that walks lineage to decide "which deals are current / which to stop targeting" (per the endpoint's own docstring, `interfaces/api/routers/deals.py:437-439`: *"The buyer's Deal Jockey can follow the lineage chain to track deal evolution"*) will never see the orphaned deal — it's neither cancelled nor discoverable, just silently dangling.

## In source

- No status guard: [`services/deal_service.py:1361-1412`](https://github.com/IABTechLab/seller-agent/blob/47beb9b4e07794d51e1e08df017f1eb67e6f335c/src/ad_seller/services/deal_service.py#L1361-L1412) — `migrate_deal` loads `old_deal`, checks only `if not old_deal` (404), then unconditionally proceeds to create a new deal and overwrite `old_deal["replacement_deal_id"]`.
- Contrast with the guard that exists one function down: [`services/deal_service.py:1440-1449`](https://github.com/IABTechLab/seller-agent/blob/47beb9b4e07794d51e1e08df017f1eb67e6f335c/src/ad_seller/services/deal_service.py#L1440-L1449) — `deprecate_deal` raises `409 already_deprecated` for the same class of "acting on an already-deprecated deal" situation.
- Router: [`interfaces/api/routers/deals.py:429-443`](https://github.com/IABTechLab/seller-agent/blob/47beb9b4e07794d51e1e08df017f1eb67e6f335c/src/ad_seller/interfaces/api/routers/deals.py#L429-L443) — no additional validation, passes straight through.
- `get_deal_lineage` (`services/deal_service.py:1470-`) is not itself buggy — its forward-walk correctly follows whatever `replacement_deal_id` chain currently exists. The corruption happens earlier, when `migrate_deal` overwrites that pointer without checking whether one was already set.

## Test coverage

Zero — there is no test file or test case anywhere in `tests/` referencing `migrate_deal`, `deprecate_deal`, or `get_deal_lineage`. The entire deal-lineage subsystem is untested.

## Suggested remedy

Mirror `deprecate_deal`'s guard in `migrate_deal`: reject (409, e.g. `already_migrated`/`already_deprecated`) when `old_deal.get("status") == "deprecated"` (or `old_deal.get("replacement_deal_id")` is already set), unless migrating a previously-migrated deal is meant to be a supported "re-migrate" flow — in which case the orphaned prior replacement should be handled explicitly (e.g. cancelled, or the new migration should chain off the *existing* replacement rather than off the original `deal_id` again) rather than silently dropped from lineage.

Happy to raise a PR once the intended behavior is confirmed — pattern-matching `deprecate_deal`'s guard is the straightforward fix if re-migration isn't meant to be allowed.

## Environment

garvitkaushik-123:
## Summary

Fixes #55. `migrate_deal` had no check on the target deal's status before mutating it. Calling `POST /api/v1/deals/{deal_id}/migrate` twice on the same `deal_id` — a duplicate client call, a retry, an operator mistake — silently created a second replacement deal and overwrote the original deal's `replacement_deal_id` to point at it. The first replacement stayed fully `confirmed` and active in storage but became permanently unreachable from `get_deal_lineage`: no incoming or outgoing pointer, just dangling.

`deprecate_deal` already guards against the same class of problem — acting again on a deal that's already deprecated — with a `409 already_deprecated` error. `migrate_deal` does something strictly *more* mutating (creates a new deal **and** deprecates the old one) and had no equivalent guard, which is what let this through. See #55 for the full repro.

## Fix

Same check `migrate_deal`'s sibling already has: if the deal is already deprecated, `409` with the existing `replacement_deal_id` so the caller isn't left guessing where the prior migration went.

```python
if old_deal.get("status") == "deprecated":
    raise HTTPException(
        status_code=409,
        detail={
            "error": "already_deprecated",
            "message": f"Deal '{deal_id}' is already deprecated and cannot be migrated again.",
            "deprecated_at": old_deal.get("deprecated_at"),
            "replacement_deal_id": old_deal.get("replacement_deal_id"),
        },
    )
```

## Test plan

`migrate_deal`, `deprecate_deal`, and `get_deal_lineage` had **zero** test coverage before this. Added `tests/unit/test_deal_lineage.py`:

- happy path for all three functions
- the double-deprecate guard that already existed (regression guard)
- the new double-migrate guard — the exact bug from #55
- a legitimate two-hop migration chain (A→B, then B→C — migrating the *current* deal each time, never re-migrating an already-deprecated one) to confirm lineage still walks correctly forward and backward after the fix

Full suite: 1408 passed (1400 + 8 new), no regressions. `ruff check` / `format --check` clean.

Closes #55

### DIFF
--- a/src/ad_seller/services/deal_service.py
+++ b/src/ad_seller/services/deal_service.py
@@ -1372,6 +1372,23 @@ async def migrate_deal(deal_id: str, request: Any) -> dict[str, Any]:
             detail={"error": "deal_not_found", "message": f"Deal '{deal_id}' not found."},
         )
 
+    # A deal that's already been migrated (or deprecated some other way)
+    # can't be migrated again — the code below would happily overwrite
+    # replacement_deal_id with a brand new deal, and the deal it had
+    # previously pointed to would still be sitting there fully active but
+    # unreachable from lineage. Same guard deprecate_deal already has for
+    # the same reason, one function down.
+    if old_deal.get("status") == "deprecated":
+        raise HTTPException(
+            status_code=409,
+            detail={
+                "error": "already_deprecated",
+                "message": f"Deal '{deal_id}' is already deprecated and cannot be migrated again.",
+                "deprecated_at": old_deal.get("deprecated_at"),
+                "replacement_deal_id": old_deal.get("replacement_deal_id"),
+            },
+        )
+
     # Build new deal from old deal + overrides
     now = datetime.utcnow()
     new_deal_id = f"DEMO-{uuid.uuid4().hex[:12].upper()}"
--- a/tests/unit/test_deal_lineage.py
+++ b/tests/unit/test_deal_lineage.py
@@ -1,0 +1,195 @@
+# Author: Green Mountain Systems AI Inc.
+# Donated to IAB Tech Lab
+
+"""Unit tests for deal migration, deprecation, and lineage (issue #55).
+
+migrate_deal, deprecate_deal, and get_deal_lineage had zero test coverage
+before this file. That's how a migrate_deal double-call went unnoticed:
+it has no guard against acting on an already-deprecated deal, unlike its
+sibling deprecate_deal, so calling /migrate twice on the same deal_id
+silently orphaned the first replacement and forked the lineage chain.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock
+
+import pytest
+from fastapi import HTTPException
+
+from ad_seller.services import deal_service
+
+
+@pytest.fixture
+def mock_storage():
+    store = {}
+    storage = AsyncMock()
+    storage.get_deal = AsyncMock(side_effect=lambda did: store.get(did))
+    storage.set_deal = AsyncMock(side_effect=lambda did, data: store.__setitem__(did, data))
+    storage._store = store
+    return storage
+
+
+def _make_deal(**overrides):
+    defaults = {
+        "deal_id": "DEAL-ORIG",
+        "deal_type": "PD",
+        "status": "confirmed",
+        "product_id": "ctv-premium-sports",
+        "actual_price_cpm": 30.0,
+        "impressions": 1_000_000,
+    }
+    defaults.update(overrides)
+    return defaults
+
+
+def _migration_request(**overrides):
+    defaults = dict(
+        deal_type=None,
+        product_id=None,
+        max_cpm=None,
+        impressions=None,
+        flight_start=None,
+        flight_end=None,
+        buyer_seat_ids=None,
+        reason="better supply path",
+    )
+    defaults.update(overrides)
+    return SimpleNamespace(**defaults)
+
+
+def _deprecation_request(**overrides):
+    defaults = dict(reason="inventory retired", replacement_deal_id=None)
+    defaults.update(overrides)
+    return SimpleNamespace(**defaults)
+
+
+class TestMigrateDeal:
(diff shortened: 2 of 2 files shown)


atc964 (approved):
Reproduced the fork on main: second migrate overwrites the original's replacement pointer while the first replacement stays fully confirmed and active — a live deal invisible to lineage. Your guard closes it, the 409 with already_deprecated mirrors the sibling deprecate_deal semantics (and returning replacement_deal_id in the detail is a nice recovery affordance the sibling lacks), and the post-conditions in the test pin the money-path properly. Full suite 1411/0 here, CI green, merge-clean.

Two pre-existing adjacencies we'll track separately, no action needed from you: cancelled deals (status=cancelled via bulk ops) can still be migrated past this guard — same hole exists in deprecate_deal; and the read-check-write race is the same storage-layer TOCTOU pattern as elsewhere. Merging in the next batch window.

PR #58, opened 2026-08-11
fix: 404 an unknown product_id in curated deals, don't fabricate a price

### LINKED ISSUE #57: create_curated_deal silently prices a nonexistent product_id at a hardcoded $12 CPM default instead of 404 — confirmed, bookable deal minted at fabricated price
## Summary

`POST /api/v1/deals/curated` (no auth dependency at all — open to anyone) accepts an arbitrary `product_id` string with no existence check against the catalog. If the supplied `product_id` doesn't match any real product — a typo, a stale reference, a deliberately made-up value — the service silently falls back to a hardcoded `base_cpm = 12.0` default and **mints and persists a fully `confirmed`, bookable deal** at that fabricated price. No 404, no validation error, nothing.

This directly contradicts the function's own docstring/comment: *"Product base price resolution... A known-but-unpriced product (no base/floor CPM) is a 422 — never a fabricated price."* The code correctly implements that guarantee for a product that exists but lacks pricing (via `catalog_service.priceable_cpm`, which does raise 422). It does **not** implement it for a product that doesn't exist in the catalog at all — that case skips price resolution entirely and lands on the hardcoded default, silently.

## Impact

Worse than a mispriced quote: this endpoint creates a **confirmed deal record** directly (`status: "confirmed"`, persisted via `storage.set_deal`), not a quote awaiting booking. There's no auth gate on the endpoint (`interfaces/api/routers/deals.py:413-414`, no `Depends(...)` at all — contrast with every other deal-creation surface). Any caller who mistypes, guesses, or deliberately supplies a bogus `product_id` gets a live deal priced at $12 CPM regardless of what the real inventory is actually worth.

## Reproduction

```python
catalog = {
    "products": {
        "ctv-premium-sports": SimpleNamespace(product_id="ctv-premium-sports", base_cpm=45.0, floor_cpm=35.0),
    }
}

request = SimpleNamespace(
    curator_id="agent-range",
    deal_type="PMP",
    product_id="ctv-premiumm-sports",  # typo -- one extra 'm', does not exist
    audience_segments=[], content_categories=[],
    impressions=1_000_000, max_cpm=None,
    flight_start=None, flight_end=None, buyer_seat_ids=[],
)

result = await deal_service.create_curated_deal(request, catalog)
```

Output:

```python
{
    'deal_id': 'CUR-128A13A0B2D5',
    'status': 'confirmed',
    'deal_type': 'PMP',
    'pricing': {'base_cpm': 12.0, 'curator_fee_cpm': 1.2, 'total_cpm': 13.2, 'currency': 'USD'},
    ...
}
```

The real product in the catalog is priced at $45 base / $35 floor. The typo'd request instead confirms a live deal at **$13.20 total CPM** — roughly a quarter of the real floor — with no error surfaced anywhere.

## In source

[`src/ad_seller/services/deal_service.py:1235-1244`](https://github.com/IABTechLab/seller-agent/blob/47beb9b4e07794d51e1e08df017f1eb67e6f335c/src/ad_seller/services/deal_service.py#L1235-L1244):

```python
# Get base price from product catalog. A known-but-unpriced product
# (no base/floor CPM) is a 422 — never a fabricated price.
base_cpm = 12.0  # Default
if request.product_id:
    product = catalog["products"].get(request.product_id)
    if product:
        from . import catalog_service
        base_cpm = catalog_service.priceable_cpm(product)
```

`catalog["products"].get(request.product_id)` returns `None` for an unknown id; the `if product:` guard then silently skips straight past pricing resolution, leaving `base_cpm` at the top-of-function default. `priceable_cpm` — the function that actually implements the "422, never fabricated" policy the comment describes (`services/catalog_service.py:441-459`) — is never even called in this path, because it's only reachable when `product` is truthy.

Compare `quotes.py`'s handling of the identical situation, which does the right thing:

```python
# routers/quotes.py — product resolution goes through quote_service.create_quote,
# which raises 404 product_not_found for an unknown product_id.
```

`create_curated_deal` has no equivalent.

The endpoint itself has no auth dependency: [`interfaces/api/routers/deals.py:413-414`](https://github.com/IABTechLab/seller-agent/blob/47beb9b4e07794d51e1e08df017f1eb67e6f335c/src/ad_seller/interfaces/api/routers/deals.py#L413-L414).

## Test coverage

Zero — `create_curated_deal` has no references anywhere in `tests/`.

## Suggested remedy

When `request.product_id` is provided but not found in the catalog, raise `404 product_not_found` (matching the pattern already used by `quote_service.create_quote` and elsewhere), the same way it's already handled for `curator_not_found`/`curator_inactive` immediately above this block. The `base_cpm = 12.0` default should only apply when no `product_id` is supplied at all (a genuinely product-less/generic curated deal, if that's an intended use case) — never as a silent substitute for "the product you asked for doesn't exist."

Happy to raise a PR for this.

## Environment

garvitkaushik-123:
## Summary

Fixes #57. `create_curated_deal`'s comment says *"A known-but-unpriced product... is a 422 — never a fabricated price,"* and the code delivers on that for a product that exists but has no `base_cpm`/`floor_cpm` (via `catalog_service.priceable_cpm`). It didn't deliver on it for a `product_id` that doesn't exist in the catalog at all: that case skipped price resolution entirely and fell through to the hardcoded `$12` CPM default meant only for the no-`product_id`-supplied case, then **minted and persisted a `confirmed` deal at that fabricated price**. No error, anywhere.

`POST /api/v1/deals/curated` has no auth dependency at all, so this was reachable by anyone — a typo, a stale product reference, a deliberately made-up `product_id` — and unlike a mispriced quote, this endpoint creates a live confirmed deal record directly, not something pending booking. See #57 for the full repro (typo'd product_id against a $45/$35 real product → confirmed deal at $13.20 total CPM).

## Fix

When `product_id` is supplied but not found in the catalog, `404 product_not_found` — the same error shape `quote_service.create_quote` already uses for the identical situation. The `$12` CPM default now only applies when no `product_id` is given at all, which the code was always treating as a separate, legitimate case (a generic curated deal with no specific product reference).

```python
base_cpm = 12.0  # Default when no product_id is supplied at all
if request.product_id:
    product = catalog["products"].get(request.product_id)
    if not product:
        raise HTTPException(
            status_code=404,
            detail={"error": "product_not_found", "message": f"Product '{request.product_id}' not found in catalog."},
        )
    from . import catalog_service
    base_cpm = catalog_service.priceable_cpm(product)
```

## Test plan

`create_curated_deal` had **zero** test coverage before this. Added `tests/unit/test_curated_deal.py`:

- unknown `product_id` → 404, no deal minted (the bug)
- known product prices off its real `base_cpm` + curator fee
- known-but-unpriced product → 422 (unchanged by this fix, confirms the existing guard still works)
- no `product_id` at all → generic $12 default (unchanged, confirms this legitimate case wasn't broken)
- `curator_not_found` → 404, for completeness

Full suite: 1408 passed, no regressions. `ruff check` / `format --check` clean.

Closes #57

### DIFF
--- a/src/ad_seller/services/deal_service.py
+++ b/src/ad_seller/services/deal_service.py
@@ -1233,14 +1233,25 @@ async def create_curated_deal(request: Any, catalog: dict[str, Any]) -> dict[str
         )
 
     # Get base price from product catalog. A known-but-unpriced product
-    # (no base/floor CPM) is a 422 — never a fabricated price.
-    base_cpm = 12.0  # Default
+    # (no base/floor CPM) is a 422 — never a fabricated price. A product_id
+    # that doesn't exist in the catalog at all is a 404, for the same
+    # reason — falling through to the no-product default here would price
+    # a typo'd/stale product_id at the generic $12 CPM instead of erroring.
+    base_cpm = 12.0  # Default when no product_id is supplied at all
     if request.product_id:
         product = catalog["products"].get(request.product_id)
-        if product:
-            from . import catalog_service
+        if not product:
+            raise HTTPException(
+                status_code=404,
+                detail={
+                    "error": "product_not_found",
+                    "message": f"Product '{request.product_id}' not found in catalog.",
+                },
+            )
 
-            base_cpm = catalog_service.priceable_cpm(product)
+        from . import catalog_service
+
+        base_cpm = catalog_service.priceable_cpm(product)
 
     # Calculate curated pricing
     curated_deal = registry.create_curated_deal(
--- a/tests/unit/test_curated_deal.py
+++ b/tests/unit/test_curated_deal.py
@@ -1,0 +1,131 @@
+# Author: Green Mountain Systems AI Inc.
+# Donated to IAB Tech Lab
+
+"""Unit tests for create_curated_deal (issue #57).
+
+create_curated_deal had zero test coverage before this file. That's how
+an unknown product_id went unnoticed silently falling back to a
+hardcoded $12 CPM default and minting a live, "confirmed" deal instead
+of 404ing -- the same honest-pricing guarantee the adjacent code comment
+claims to enforce, just missing for the "product doesn't exist at all"
+case rather than the "product exists but has no price" case it does
+cover.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock
+
+import pytest
+from fastapi import HTTPException
+
+from ad_seller.services import deal_service
+
+
+@pytest.fixture
+def mock_storage():
+    store = {}
+    storage = AsyncMock()
+    storage.set_deal = AsyncMock(side_effect=lambda did, data: store.__setitem__(did, data))
+    storage._store = store
+    return storage
+
+
+def _make_product(**overrides):
+    defaults = dict(product_id="ctv-premium-sports", base_cpm=45.0, floor_cpm=35.0)
+    defaults.update(overrides)
+    return SimpleNamespace(**defaults)
+
+
+def _catalog(*products):
+    return {"products": {p.product_id: p for p in products}}
+
+
+def _curated_request(**overrides):
+    defaults = dict(
+        curator_id="agent-range",
+        deal_type="PMP",
+        product_id=None,
+        audience_segments=[],
+        content_categories=[],
+        impressions=1_000_000,
+        max_cpm=None,
+        flight_start=None,
+        flight_end=None,
+        buyer_seat_ids=[],
+    )
+    defaults.update(overrides)
+    return SimpleNamespace(**defaults)
+
+
+class TestCreateCuratedDeal:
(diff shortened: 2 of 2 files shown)


atc964 (approved):
Reproduced on main exactly per #57: unknown product_id in a curated deal fell through to the $12 default and persisted a confirmed deal at a fabricated price — the worst variant of this class since nothing downstream re-verifies curated deals. Your 404 lands before registry/storage writes (no partial state), preserves the legitimate no-product_id default, and matches the existing product_not_found shape byte-for-byte. Bug test fails on main, the other four are honest regression guards, suite 1408/0 here, CI green.

Completeness sweep on our side confirms every other product-pricing path already 404s unknowns — this was the last hole in the fabricated-price class. A few adjacent items of a different class (migrate_deal product refs, zero-CPM negotiation cold path, a 422 shape inconsistency) are pre-existing and tracked internally. Merging in the next batch window — thanks for another precise one.

PR #59, opened 2026-08-12
Cleans up operator key management

therevoltingx:
### Problem

`ad-seller create-operator-key` had two rough edges after the operator-auth work landed:

1. **Duplicate labels** — creating another operator key with an existing active label succeeded silently, making it hard to tell which key was which.
2. **Process hang** — the command printed the key but never exited. Storage was never closed after `asyncio.run(...)`, and the console-script entrypoint (`ad-seller = ...:app`) skipped the hard `os._exit` that `__main__` used to bypass telemetry atexit hangs.

Operators also had no out-of-band way to **list** or **revoke** operator keys without going through the HTTP API (which itself requires an operator credential).

### What this PR does

**Duplicate active labels**

- `ApiKeyService.create_operator_key` rejects an active (non-revoked, non-expired) operator key with the same label (`ValueError`, includes existing `key_id`).
- `POST /auth/api-keys/operator` maps that to **409 Conflict**.
- Revoked/expired labels may be reused. Buyer keys are unaffected.

**CLI exits cleanly**

- `create-operator-key` (and the new commands) always `close_storage()` in a `finally`.
- Console script entrypoint is now `ad_seller.interfaces.cli.main:main`, wrapped with `force_exit_after` so `ad-seller` hard-exits like `__main__`.

**Bootstrap CLI surface**

| Command | Purpose |
|---------|---------|
| `ad-seller create-operator-key` | Mint (unchanged, plus label uniqueness) |
| `ad-seller list-operator-keys [--include-inactive]` | List operator keys (metadata only) |
| `ad-seller delete-operator-key --label …` / `--key-id …` | Soft-revoke an operator key; frees the label |

`delete-operator-key` refuses buyer keys (use `DELETE /auth/api-keys/{key_id}` for those).

### Docs

Updated bootstrap sections in `docs/api/authentication.md`, `docs/guides/agent-management.md`, and `docs/guides/developer-setup.md`.

### Test plan

- [ ] `ad-seller create-operator-key --label "Ops"` twice → second fails with clear error; process exits
- [ ] `ad-seller list-operator-keys` shows the active key; `--include-inactive` shows revoked too
- [ ] `ad-seller delete-operator-key --label "Ops"` then recreate with the same label succeeds
- [ ] `POST /auth/api-keys/operator` with a duplicate active label returns 409
- [ ] Reinstall entrypoint if needed: `pip install -e .`
- [ ] `pytest tests/unit/test_operator_auth.py`

### DIFF
--- a/docs/api/authentication.md
+++ b/docs/api/authentication.md
@@ -41,6 +41,21 @@ ad-seller create-operator-key --label "Primary operator"
 
 Run this with the same storage config (`.env`) as the server so the key lands in the backend the server reads. The full key is printed **once** — store it securely.
 
+List operator keys (metadata only — secrets are never re-shown):
+
+```bash
+ad-seller list-operator-keys
+ad-seller list-operator-keys --include-inactive   # include revoked/expired
+```
+
+To revoke an operator key out-of-band (frees the label for reuse):
+
+```bash
+ad-seller delete-operator-key --label "Primary operator"
+# or
+ad-seller delete-operator-key --key-id key-a1b2c3d4
+```
+
 Subsequent operator keys can be minted over HTTP with an existing operator credential (see below).
 
 ## Operator Surface
--- a/docs/guides/agent-management.md
+++ b/docs/guides/agent-management.md
@@ -21,6 +21,14 @@ the first operator key with the CLI (no HTTP auth):
 ad-seller create-operator-key --label "Primary operator"
 ```
 
+List and revoke the same way:
+
+```bash
+ad-seller list-operator-keys
+ad-seller delete-operator-key --label "Primary operator"
+# or: ad-seller delete-operator-key --key-id key-a1b2c3d4
+```
+
 See [Authentication](../api/authentication.md) for the full operator surface.
 
 ### Create a Buyer API Key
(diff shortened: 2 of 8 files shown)


atc964 (approved):
Verified all three fixes, including reproducing the hang mechanism empirically: aiosqlite's connection is a non-daemon thread, so an unclosed storage after asyncio.run blocks interpreter shutdown — the close_storage-in-finally is the root-cause fix and routing the console script through force_exit_after is sound insurance (rich flushes per print, so the one-time key survives the hard exit even piped). Duplicate-label 409 confirmed at service, HTTP, and CLI layers with an end-to-end smoke on the installed entrypoint; no weakening of the operator-auth surface (deps untouched, no new routes; the 409's key_id disclosure is operator-visible anyway). Suite 1421/0 here.

Three take-or-leave observations: (1) label uniqueness is check-then-mint, so two concurrent operator POSTs with the same label can both land — low stakes here, maybe worth a docstring line; (2) the blanket ValueError->409 in the router will silently 409 any future ValueError from that path — a dedicated exception would be tighter; (3) behavior note: a second unlabeled operator-key request now 409s on the empty/default label where it used to succeed — reads as intended, but flagging since it changes bootstrap scripts that mint twice.

Editable installs need pip install -e . re-run for the entrypoint change, as your body notes. Good cleanup.

Issue #60, opened 2026-08-18
Stale '@listen() CrewAI version mismatch' test stub hides zero coverage on discovery_inquiry_flow (execution_activation_flow has a real, different bug)


garvitkaushik-123:
## Summary

Every test file that touches `ad_seller.flows` (dozens of them) carries this boilerplate, first introduced in `530df34` (March 2026):

```python
# Stub broken flow modules (pre-existing @listen() bugs with CrewAI version mismatch)
_broken_flows = [
    "ad_seller.flows.discovery_inquiry_flow",
    "ad_seller.flows.execution_activation_flow",
]
```

It replaces both modules with fake empty classes before any test runs. I went looking for something else and ran into this, then actually tried running both flows for real instead of trusting the comment. The two modules aren't in the same shape — I want to report them separately rather than as one blob.

## Part 1 — `discovery_inquiry_flow.py`: the stated reason is stale

At `530df34`, `crewai` was pinned at `>=0.86.0` (pre-1.0). It's `>=1.14.4,<2.0.0` now (resolves to `1.15.2`) — a major version bump, the kind that plausibly broke and later fixed a decorator API like `@listen()`. So I tested it directly rather than guessing:

- `DiscoveryInquiryFlow().kickoff_async(...)` completes cleanly across all four routing branches (default/catalog, `"price"`→pricing, `"available"`→availability, `"target"`→targeting) — every listener finishes, `state.status == COMPLETED` every time.
- Verified both as a standalone script and under `pytest` itself (the actual context the stub is guarding), with a minimal isolated test file — no framework error anywhere.
- Also isolated the specific `or_()` combinator pattern `execution_activation_flow.py` uses (see Part 2) in a bare-bones `Flow` against this same pinned `crewai` version — works correctly.

None of that reproduces a `@listen()` bug. As a direct consequence, `grep -rl "DiscoveryInquiryFlow" tests/` returns nothing — the stub means no test file has ever exercised this module's real code, so nobody would notice either way. 295 lines of production flow logic with zero coverage.

## Part 2 — `execution_activation_flow.py`: a real bug, just not this one

This one *does* crash for me, but not from `@listen()`. Running `ExecutionActivationFlow().kickoff_async(...)` (deal_id path, no live ad server configured) blows up with:

```
Task exception was never retrieved
future: <Task finished name='Task-7' ...> exception=RuntimeError('Attempted to exit cancel scope in a different task than it was entered in')>
```

[...]

PR #63, opened 2026-08-19
test: stop stubbing discovery_inquiry_flow, restore real coverage

garvitkaushik-123:
## Summary

Part 1 of #60 (green-lit by @atc964 on the issue).

The `"pre-existing @listen() bugs with CrewAI version mismatch"` stub that every `ad_seller.flows`-touching test file carried for `discovery_inquiry_flow` dates to when `crewai` was pinned at `>=0.86.0` (commit `530df34`, March 2026). It's `>=1.14.4` now (resolves to `1.15.2`) — a major version bump. Verified on the issue that the flow runs cleanly end-to-end on the current version, including the exact production call path (`flow.query()`'s sync `self.kickoff()` invoked from inside an async FastAPI handler).

This wasn't just stale test hygiene — `DiscoveryInquiryFlow` backs a live endpoint (`POST /discovery`, `products.py:201`), so the blanket stub meant that endpoint had zero real test coverage for as long as the stub existed (`grep -rl "DiscoveryInquiryFlow" tests/` returned nothing before this PR).

`execution_activation_flow` stays stubbed — it has a separate, real bug (a cancel-scope leak in an MCP client cleanup path, not the `@listen()` issue the old comment claimed) that's tracked as #60 part 2 and will be fixed independently.

## Changes

- Removed the `discovery_inquiry_flow` entry from the `_broken_flows` stub list in all 33 test files that had it. `execution_activation_flow` entries are untouched.
- Simplified `test_deal_flow_e2e.py`: its `_get_deal_request_flow_class()` helper existed solely to bypass `flows/__init__.py` via manual `sys.modules` surgery so importing `DealRequestFlow` wouldn't also trigger `discovery_inquiry_flow`. With the stub gone that's unnecessary — replaced with a plain module import (net -41 lines in that file alone).
- Fixed the stale docstring in `test_linear_tv.py` making the same claim.
- Added `tests/unit/test_discovery_inquiry_flow.py`: the flow's own routing logic across all four response types (`catalog`/`pricing`/`availability`/`targeting`) run for real, plus `POST /discovery` exercised through the actual FastAPI app with the real flow — not mocked — closing the exact coverage gap the issue found.

## Test plan

Full suite: 1491 passed, 28 skipped (pre-existing, unrelated to this change), no regressions. `ruff check` / `format --check` clean.

Part of #60.

### DIFF
--- a/tests/integration/test_deal_flow_e2e.py
+++ b/tests/integration/test_deal_flow_e2e.py
@@ -7,9 +7,7 @@ Tests the key flow components with mocked external dependencies:
 - Approval gate integration
 """
 
-import importlib
-import sys
-
+import ad_seller.flows.deal_request_flow as deal_request_flow_module
 from ad_seller.models.buyer_identity import BuyerContext, BuyerIdentity
 from ad_seller.models.core import DealType, PricingModel
 from ad_seller.models.flow_state import (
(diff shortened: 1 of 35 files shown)


Sirajmx (approved):
Verified: The full suite matches the stated plan exactly - 1,491 passed, 28 skipped, with Ruff clean. There are no src/ changes, so there is no direct production-code behavioral impact.

Also ran end-to-end testing against a live server with no regressions across the route surface.

Looks good to merge.

PR #64, opened 2026-08-19
fix: enforce idempotency for change requests

### LINKED ISSUE #62: POST /api/v1/change-requests accepts no idempotency_key and creates duplicate change requests on replay
## Summary

`POST /api/v1/change-requests` is the fourth of four money-mutating surfaces the shared contract enumerates under FD-12. The other three implement it; this one has no `idempotency_key` field at all, so a replayed request silently creates a second change request and routes a second approval for the same logical change.

## Observed: the contract names this endpoint

`iab-agentic-primitives` v0.5.1 (`ee9ea3c`) defines the covered set explicitly in [`PROTOCOL_RECONCILIATION.md`](https://github.com/IABTechLab/iab-agentic-primitives/blob/ee9ea3c1c612b1be6ae00b75e0ab1b81435f62cc/PROTOCOL_RECONCILIATION.md), decision row P2:

> every money-mutating request inherits `IdempotentRequest` — `QuoteRequest`, `DealBookingRequest`, `NegotiationMessage`, `ChangeRequestCreate` all carry a REQUIRED `idempotency_key`

The section headed *"Idempotency replay semantics (FD-12) - normative summary"* lists `/api/v1/change-requests` alongside `/api/v1/quotes`, `/api/v1/deals` and `/api/v1/negotiations/messages`. And [`protocol/deals.py:98`](https://github.com/IABTechLab/iab-agentic-primitives/blob/ee9ea3c1c612b1be6ae00b75e0ab1b81435f62cc/src/iab_agentic_primitives/protocol/deals.py#L98) declares:

```python
class ChangeRequestCreate(IdempotentRequest):
    """Request body for ``POST /api/v1/change-requests`` (money-mutating: FD-12)."""
```

## Observed: the seller does not use it

Three of the four routers import the contract type directly and implement FD-12. The fourth does not:

| Router | Request type | FD-12 |
|---|---|---|
| [`quotes.py:18`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/interfaces/api/routers/quotes.py#L18) | `QuoteRequest` (contract) | implemented |
| [`deals.py:20`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/interfaces/api/routers/deals.py#L20) | `DealBookingRequest` (contract) | implemented |
| [`negotiation.py:20`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/interfaces/api/routers/negotiation.py#L20) | `NegotiationMessage` (contract) | implemented |
| [`change_requests.py:12`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/interfaces/api/routers/change_requests.py#L12) | `CreateChangeRequestModel` (local) | **absent** |

[`schemas.py:355`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/interfaces/api/schemas.py#L355) carries six fields — `order_id`, `change_type`, `diffs`, `proposed_values`, `reason`, `requested_by` — and no key. `idempotency_key` does not appear anywhere in `schemas.py`, and no service module implements idempotency, so it is not handled at a lower layer either.

The model does not set `extra="forbid"`, so a caller who supplies `idempotency_key` has it silently discarded rather than rejected.

## Observed: runtime

Against v2.4.2 (`e5b367d`), two identical requests carrying the same `idempotency_key`:

```text
call 1 -> CR-236B96AC3FCC  status: pending_approval  severity: material
call 2 -> CR-299478430C51  status: pending_approval  severity: material
```

Two distinct change requests, both classified `material`, both routed to approval. So a retried request does not only duplicate a record — it places two identical material changes into the seller's approval queue for one logical change, leaving a human to determine which to action.

For contrast, on the same build:

```text
POST /api/v1/quotes with no idempotency_key -> 422
```

## Reproduction

```bash
# 1. Create an order. deal_id is required here; see the note below.
ORD=$(curl -s -X POST localhost:8001/api/v1/orders \
  -H "Content-Type: application/json" -d '{"deal_id":"DEAL-PROBE-1"}' \
  | python3 -c "import json,sys; print(json.load(sys.stdin)['order_id'])")

# 2. Send the same body with the same key twice.
BODY=$(printf '{"order_id":"%s","change_type":"impressions","reason":"probe","idempotency_key":"probe-key-0001"}' "$ORD")
for i in 1 2; do
  curl -s -X POST localhost:8001/api/v1/change-requests \
    -H "Content-Type: application/json" -d "$BODY" \
    -o "/tmp/cr$i.json" -w "call $i: HTTP %{http_code}  "
  python3 -c "import json; d=json.load(open('/tmp/cr$i.json')); print(d['change_request_id'], d['status'], d['severity'])"
done
```

Observed on v2.4.2:

```text
call 1: HTTP 200  CR-236B96AC3FCC pending_approval material
call 2: HTTP 200  CR-299478430C51 pending_approval material
```

Two different `change_request_id` values from an identical body under the same key is the finding.

Note: the order must carry a `deal_id`. Creating one without it produces an unhandled 500 on this endpoint; see #61.

## Environment

seller-agent v2.4.2, commit `e5b367d2b780aa4d0e03b6363d744ba3adf2b221`, installed with `uv sync --locked`, resolving `iab-agentic-primitives==0.5.1` (`ee9ea3c`). Python 3.12.3, WSL2 Ubuntu 24.04.

Also confirmed on `main`, which is currently the same commit (`e5b367d`) at time of reporting: `change_requests.py` imports the local `CreateChangeRequestModel`, and `idempotency_key` does not appear in `schemas.py`.

## Inferred

`PROTOCOL_RECONCILIATION.md` describes the retired makegood and cancellation sub-routes as re-pointed at `/api/v1/change-requests`, with a note that the buyer's client call should stay disabled "until the seller implements change requests". A change-requests surface does exist and persists state, so it is not clear whether it predates the contract and a conformant implementation is still planned, or whether it was implemented and FD-12 was missed. The gap is the same either way, but the intended status would be useful to know.

## Suggested remedy

The pattern exists three times in this codebase, and #50 (quotes) is the closest analogue. It also left reusable helpers in `contract_mappers.py` — `idempotency_conflict_detail()`, `request_payload_hash()` and `idempotency_buyer_scope()` — and established the ordering that review on that PR found necessary: resolve and verify the caller *before* the idempotency lookup, then scope the storage key by the resulting identity (`quotes.py:71` and `:89`).

Two parts:

- Accept `idempotency_key` on the request, either by adopting the contract's `ChangeRequestCreate` or by adding the field to `CreateChangeRequestModel`.
- Replay an identical request rather than minting a new `change_request_id`, and return `idempotency_conflict` (409) where the same key arrives with a different payload.

One scoping question I do not have an answer to. `/quotes` scopes by verified buyer context, but this endpoint uses `_get_optional_api_key_record` rather than `_verified_buyer_context`, so the same verified principal is not available here — the same constraint that came up for `book_deal` on #50. Since `order_id` is required and the order already carries the buyer relationship, scoping by order may be more natural than scoping by caller identity. Flagging it rather than proposing it, as the right answer depends on how change requests are intended to be authorised.

## Acceptance criteria

- An identical `POST /api/v1/change-requests` replayed under the same `idempotency_key` returns the original change request and does not create a second approval.
- The same key with a materially different payload returns `idempotency_conflict`.
- A request omitting `idempotency_key` is handled consistently with the other three FD-12 surfaces.

garvitkaushik-123:
## Summary

- require a non-empty idempotency key on change-request creation
- replay the original change request for identical requests scoped to the same order
- return the shared idempotency conflict response when a key is reused with a different payload
- expire idempotency records after 24 hours, matching the other FD-12 endpoints

## Why

POST /api/v1/change-requests was the only FD-12 money-mutating surface without replay protection. A client retry could create a second change request and place a duplicate material change into the approval queue.

The implementation follows the existing quotes/deals router pattern while preserving the current change-request payload fields. The namespace is scoped by order because this route permits optional authentication.

## Testing

- added coverage for missing keys, identical replay, changed-payload conflicts, per-order scoping, and the 24-hour TTL
- updated existing change-request lifecycle tests for the required field
- full suite: 1,484 passed, 28 skipped
- changed-file lint, formatting, and whitespace checks pass

Closes #62

### DIFF
--- a/src/ad_seller/interfaces/api/routers/change_requests.py
+++ b/src/ad_seller/interfaces/api/routers/change_requests.py
@@ -5,9 +5,10 @@
 
 from typing import Optional
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
 
 from ....services import order_service
+from .. import contract_mappers as cm
 from .. import deps
 from ..schemas import CreateChangeRequestModel, ReviewChangeRequestModel
 
@@ -23,8 +24,51 @@ async def create_change_request(
 
     Validates the change against the current order state, classifies
     severity, and routes to approval if needed.
+
+    **Idempotency (FD-12):** the request carries a required
+    ``idempotency_key``. An identical replay returns the original change
+    request without creating a second approval; reusing the key with a
+    different body returns ``idempotency_conflict`` (HTTP 409). Keys are
+    scoped per order and expire after 24 hours.
     """
-    return await order_service.create_change_request(request)
+    from ....storage.factory import get_storage
+
+    storage_key = f"idempotency:change-request:{request.order_id}:{request.idempotency_key}"
+    payload_hash = cm.request_payload_hash(
+        request.model_dump(mode="json", exclude={"idempotency_key"})
+    )
+    storage = await get_storage()
+    try:
+        prior = await storage.get(storage_key)
+    except Exception:
+        prior = None
+
+    if isinstance(prior, dict) and prior.get("change_request_id"):
+        if prior.get("payload_hash") != payload_hash:
+            raise HTTPException(
+                status_code=409,
+                detail=cm.idempotency_conflict_detail(
+                    f"idempotency_key '{request.idempotency_key}' was already used "
+                    "for a different change request."
+                ),
+            )
+        return await order_service.get_change_request(prior["change_request_id"])
+
+    result = await order_service.create_change_request(request)
+
+    try:
+        await storage.set(
+            storage_key,
+            {
+                "change_request_id": result["change_request_id"],
+                "payload_hash": payload_hash,
+            },
+            ttl=86400,
+        )
+    except Exception:
+        pass
+
+    return result
 
 
 @router.get("/api/v1/change-requests", tags=["Change Requests"])
--- a/src/ad_seller/interfaces/api/schemas.py
+++ b/src/ad_seller/interfaces/api/schemas.py
@@ -38,7 +38,7 @@ from iab_agentic_primitives.protocol import (
 from iab_agentic_primitives.protocol import (
     ProductAvailsSearch as ProductAvailsSearch,  # noqa: PLC0414 — explicit re-export
 )
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class PricingRequest(BaseModel):
@@ -355,6 +355,7 @@ class FieldDiffModel(BaseModel):
 class CreateChangeRequestModel(BaseModel):
     """Request to create a change request for an order."""
 
+    idempotency_key: str = Field(min_length=1)
     order_id: str
     change_type: str
     diffs: list[FieldDiffModel] = []
--- a/tests/unit/test_change_request.py
+++ b/tests/unit/test_change_request.py
@@ -170,6 +170,7 @@ class TestCreateChangeRequest:
             resp = await client.post(
                 "/api/v1/change-requests",
                 json={
+                    "idempotency_key": "idem-minor-1",
                     "order_id": "ORD-TEST001",
                     "change_type": "creative",
                     "reason": "Swap banner creative",
@@ -190,6 +191,7 @@ class TestCreateChangeRequest:
             resp = await client.post(
                 "/api/v1/change-requests",
                 json={
+                    "idempotency_key": "idem-material-1",
                     "order_id": "ORD-TEST001",
                     "change_type": "impressions",
                     "diffs": [{"field": "impressions", "old_value": 5000000, "new_value": 8000000}],
@@ -208,6 +210,7 @@ class TestCreateChangeRequest:
             resp = await client.post(
                 "/api/v1/change-requests",
                 json={
+                    "idempotency_key": "idem-rollback-1",
                     "order_id": "ORD-TEST001",
                     "change_type": "flight_dates",
                     "diffs": [
@@ -228,6 +231,7 @@ class TestCreateChangeRequest:
             resp = await client.post(
                 "/api/v1/change-requests",
                 json={
+                    "idempotency_key": "idem-missing-order-1",
                     "order_id": "ORD-NOPE",
                     "change_type": "creative",
                 },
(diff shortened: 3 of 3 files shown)


Sirajmx (commented):
> Confirmed, and the underlying fix here is solid - reviewed it against #62's replay/duplication
> scenario and it holds. But the hygiene gate is genuinely red right now
> (`hygiene / secret scan (gitleaks, changed range)`), so this can't merge as-is.
>
> Ran gitleaks 8.24.3 locally, unredacted, against the same range CI checks
> (`e5b367d..2c2633d`). It's a false positive, not a real leak:
>
> ```
> Finding:  "idempotency_key": "idem-apply-unapproved-1"
> RuleID:   generic-api-key    Entropy: 3.708132
> File:     tests/unit/test_change_request.py:601
> ```
>
> `idem-apply-unapproved-1` is a plain test fixture in `test_cannot_apply_unapproved` - same
> pattern as `idem-minor-1`, `idem-material-1`, etc. elsewhere in this file. It just has enough
> entropy in the longer name to trip gitleaks' generic-api-key rule where the shorter ones don't.
>
> Direction: add an inline `# gitleaks:allow` comment on that line. One thing worth knowing before
> you push it — gitleaks scans each commit in the range individually, not a squashed diff, so
> adding a *new* commit on top won't clear the gate; the original commit still carries the
> flagged line in history. **Amend the existing commit** (`git commit --amend`) and force-push,
> rather than stacking a fix commit. Verified locally both ways: a stacked commit still shows
> `leaks found: 1` over the full range; amending clears it to `no leaks found`.

garvitkaushik-123:
Gitleaks false positive fixed (inline allow comment, commit amended and force-pushed). All checks are green now, including the secret scan.

This is ready to merge but needs a formal approving review. @Sirajmx, could you approve when you get a chance?

Sirajmx (approved):
Verified: 1,463 unit + 39 integration tests passed (0 failed, matching CI exactly). Live E2E also passed - identical replays return the same record, while reusing a key with a different body correctly returns 409. Merging.

PR #65, opened 2026-08-19
fix: reject change requests for dealless orders

### LINKED ISSUE #61: POST /api/v1/change-requests returns 500 for an order created without a deal_id
## Summary

`POST /api/v1/orders` allows an order to be created without a `deal_id`, storing the key with value `None`. A subsequent `POST /api/v1/change-requests` for that order raises an unhandled `ValidationError` and returns an unstructured HTTP 500 (`Internal Server Error`) — despite the handler having explicit 400, 404 and 422 paths for every other failure it anticipates.

## Observed

`CreateOrderRequest` ([`schemas.py:332`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/interfaces/api/schemas.py#L332)) declares `deal_id`, `quote_id` and `metadata` all `Optional`, and `create_order` persists the value as given ([`order_service.py:46`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/services/order_service.py#L46)):

```python
order_data["deal_id"] = deal_id
```

So an order created without a `deal_id` is stored with the key present and the value `None`.

`create_change_request` then reads it ([`order_service.py:348`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/services/order_service.py#L348)):

```python
deal_id=order.get("deal_id", ""),
```

The `""` default applies only when the key is **absent**. Here the key is present with value `None`, so `.get()` returns `None`, and [`ChangeRequest.deal_id` is typed `str`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/models/change_request.py#L64) — its `= ""` default does not apply to an explicitly passed `None`.
```text
pydantic_core._pydantic_core.ValidationError: 1 validation error for ChangeRequest
deal_id
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

The request is well formed and the order legitimately created through the service's own public endpoint. The handler already returns structured errors for an invalid `change_type` (400), an unknown order (404) and failed state validation (422); this path returns an unstructured `Internal Server Error` rather than a documented error.

## Reproduction

```bash
ORD=$(curl -s -X POST localhost:8001/api/v1/orders \
  -H "Content-Type: application/json" -d '{}' \
  | python3 -c "import json,sys; print(json.load(sys.stdin)['order_id'])")

curl -s -w "\nHTTP %{http_code}\n" -X POST localhost:8001/api/v1/change-requests \
  -H "Content-Type: application/json" \
  -d "{\"order_id\":\"$ORD\",\"change_type\":\"impressions\",\"reason\":\"probe\"}"
```

Returns `Internal Server Error` / HTTP 500. Supplying `{"deal_id":"DEAL-PROBE-1"}` at step one succeeds, which isolates the null value as the cause.

## Environment

seller-agent v2.4.2, commit `e5b367d2b780aa4d0e03b6363d744ba3adf2b221`, installed with `uv sync --locked`. Python 3.12.3, WSL2 Ubuntu 24.04.

## Inferred

Not established, and a design call rather than something the evidence settles: whether the correct fix is at the write side (reject or normalise a null `deal_id` in `create_order`), at the read side (`order.get("deal_id") or ""`), or at the model (`Optional[str]`). The read-side change is the smallest and matches the evident intent of the existing `""` default, but it leaves orders with null `deal_id` in storage, which may or may not be wanted.

Also unverified: whether other consumers of `storage.get_order` make the same assumption. I have not swept for it.

## Suggested remedy

One minimal defensive fix is `order.get("deal_id") or ""` at `order_service.py:348`, which restores the behaviour the existing `""` default was reaching for, but only if an empty `deal_id` is a valid representation for a change request. If it is not, this would convert a 500 into a successfully created change request carrying an empty deal reference, which may conceal an invalid lifecycle state; a documented 4xx would then be the better outcome.

If a null `deal_id` should not be persisted at all, normalising or rejecting it in `create_order` is the more thorough fix, though it would leave existing stored orders unaffected.

## Acceptance criteria

- For an order created through `POST /api/v1/orders` without a `deal_id`, `POST /api/v1/change-requests` either creates a semantically valid change request or returns a documented client error. It does not raise an uncaught exception or return HTTP 500.

garvitkaushik-123:
## Summary

- normalize omitted order deal IDs to an empty string instead of persisting null
- reject change requests for orders without a deal reference using a structured HTTP 400 response
- guard legacy records where deal_id is already null before constructing the change-request model

## Why

Orders created without a deal ID are valid through the public orders endpoint, but requesting a change against one passed null into the ChangeRequest model and raised an uncaught validation error. That surfaced as an unstructured HTTP 500.

The write-side normalization prevents new null records, while the read-side check protects records already stored with the old shape.

## Testing

- verifies new dealless orders persist an empty deal ID
- verifies a legacy order with a null deal ID returns deal_id_required with HTTP 400
- verifies the rejected request does not create a change-request record
- full suite: 1,480 passed, 28 skipped
- changed-file lint, formatting, and whitespace checks pass

Closes #61


Sirajmx (commented):
> Confirmed the fix, and it's correct on its own — verified `test_existing_order_with_null_deal_id_returns_structured_error`
> passes in isolation. But there's a real interaction with #64 worth flagging before this merges,
> since the two of you opened within 10 minutes of each other and neither could have seen the
> other's change.
>
> #64 makes `idempotency_key` a required field on `POST /api/v1/change-requests`. Your new test's
> payload doesn't include one:
>
> ```python
> json={
>     "order_id": "ORD-TEST001",
>     "change_type": "impressions",
>     "reason": "Increase campaign reach",
> },
> ```
>
> Once #64 is in, that request gets rejected at validation (422) before it ever reaches the
> `deal_id_required` check this test exists to verify — so the test fails on 422 vs the expected
> 400. Confirmed this locally, deterministically, merging both onto a common base.
>
> Direction: add `"idempotency_key": "idem-null-deal-1"` (or any non-empty string) to that one
> payload. That's the only call this PR makes to that endpoint, so it's the only place this bites.
> Verified the one-line fix locally — full suite green (1498 passed, 0 failed) with both PRs and
> #63 combined. Happy to merge right after #64 once this is in; no reason to block on ordering
> otherwise.

atc964:
Checking in on this. Siraj verified the fix and found only one interaction with the change-request idempotency change that merged after you opened this: the new test payload needs an idempotency_key (his comment has the exact line). If you can push that line we will merge right away; if you are busy, say the word and we will apply it as a maintainer edit with credit to you by Thursday.

garvitkaushik-123:
Pushed @atc964, thanks for catching this and verifying it locally.

Added \`"idempotency_key": "idem-null-deal-1"\` to that one payload. Rebased onto current main (now includes #64) rather than just adding the line in place, so the test actually runs against the merged idempotency requirement rather than my own guess at it.

\`test_existing_order_with_null_deal_id_returns_structured_error\` passes, full suite: 1503 passed, 0 failed, 28 skipped (pre-existing, unrelated). \`ruff check\` / \`format --check\` clean on the changed file.

Ready for merge whenever convenient.

Sirajmx (approved):
> Verified: full unit + integration suite passes, 0 failed, merged onto current main. Live e2e
> confirms it too - a legacy null-deal_id change-request genuinely 500s on main, 400
> `deal_id_required` on this branch; new orders now persist `deal_id: ""` not `null`. Merging.

Issue #69, opened 2026-08-22
Rate card is writable but is not read by the quote, deal-booking or negotiation pricing paths


jaanijuk:
## Summary

`GET`/`PUT /api/v1/rate-card` store and return a rate card document under the `Pricing` tag. Quotes, deal booking and negotiation all take their base price from the product catalogue instead, so changing the rate card has no effect on what the seller quotes or accepts. Neither the rate-card endpoints nor the quote rationale indicate this.

## Observed — a stored rate card does not change a quote

Starting state, six entries:

```text
display 12.00 | video 25.00 | ctv 35.00 | mobile_app 18.00 | native 10.00 | audio 15.00
```

The product under test. Note that the identifier and name suggest CTV, while the catalogue declares the inventory type as `display` — so `display` is the rate-card key that applies:

```bash
curl -s localhost:8001/products | python3 -c "
import json,sys
p=json.load(sys.stdin); p=p.get('products',p)
x=[i for i in p if i['product_id']=='inv-ctv-apex-series'][0]
print(x['ext']['inventory_type'], x['base_price']['amount_micros']/1e6, x['ext']['floor_cpm'])
"
# display 45.0 38.25
```

Set `display` to 99.00, leaving the other five untouched. The request body is a bare array of entries:

```bash
curl -s -X PUT localhost:8001/api/v1/rate-card \
  -H "Authorization: Bearer $OPERATOR_KEY" -H "Content-Type: application/json" \
  -d '[{"inventory_type":"display","base_cpm":99.0,"currency":"USD"},
       {"inventory_type":"video","base_cpm":25.0,"currency":"USD"},
       {"inventory_type":"ctv","base_cpm":35.0,"currency":"USD"},
       {"inventory_type":"mobile_app","base_cpm":18.0,"currency":"USD"},
       {"inventory_type":"native","base_cpm":10.0,"currency":"USD"},
       {"inventory_type":"audio","base_cpm":15.0,"currency":"USD"}]'
# HTTP 200
```

Read back to confirm it stored:

```text
stored: [('display', 99.0), ('video', 25.0), ('ctv', 35.0), ('mobile_app', 18.0), ('native', 10.0), ('audio', 15.0)]
```

Quote a product whose `inventory_type` is `display`:

```bash
curl -s -X POST localhost:8001/api/v1/quotes -H "Content-Type: application/json" \
[...]

PR #72, opened 2026-08-31
fix: hydrate approval resume via Flow._state, not the read-only property

### LINKED ISSUE #67: POST /approvals/{approval_id}/resume returns 500 for both approved and rejected decisions: assignment to read-only ProposalHandlingFlow.state
## Summary

`POST /approvals/{approval_id}/resume` is the documented mechanism for completing a proposal-handling flow after a human decision. It returned an unstructured HTTP 500 for both decision types tested, because `_resume_proposal_flow` assigns to `ProposalHandlingFlow.state`, which is a read-only property in the lock-resolved CrewAI 1.15.2.

Approval decisions themselves are unaffected: they persist correctly, stamp the verified principal, and emit the expected events. What fails is the step that applies the decision to the flow.

## Observed

Against v2.4.2 (`e5b367d`), with `APPROVAL_GATE_ENABLED=true` and `APPROVAL_REQUIRED_FLOWS=proposal_decision`, two gated proposals were created and decided — one approved, one rejected. Both decisions succeeded. Both resume calls returned:

```text
HTTP 500
Internal Server Error
```

Server traceback, identical for both:

```text
File "src/ad_seller/interfaces/api/routers/approvals.py", line 78, in resume_flow
  return await approval_service.resume_flow(approval_id)
File "src/ad_seller/services/approval_service.py", line 108, in resume_flow
  return await _resume_proposal_flow(request, response)
File "src/ad_seller/services/approval_service.py", line 129, in _resume_proposal_flow
  flow.state = ProposalState(**snapshot)
File ".venv/lib/python3.12/site-packages/crewai/flow/runtime/__init__.py", line 747, in __setattr__
  object.__setattr__(self, name, value)
AttributeError: property 'state' of 'ProposalHandlingFlow' object has no setter
```

The line in question, [`approval_service.py:128-129`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/services/approval_service.py#L128-L129):

```python
flow = ProposalHandlingFlow()
flow.state = ProposalState(**snapshot)
```

This is the only assignment to `.state` on a flow object anywhere in `src/`.

## Observed — the endpoint is the documented completion path

The router docstring states its purpose plainly ([`approvals.py:72-77`](https://github.com/IABTechLab/seller-agent/blob/e5b367d2b780aa4d0e03b6363d744ba3adf2b221/src/ad_seller/interfaces/api/routers/approvals.py#L72-L77)):

> Loads the flow state snapshot, applies the decision, and returns the final result without re-running expensive crew evaluations.

`_resume_proposal_flow` is where the decision is actually applied — appending to `accepted_proposals` or `rejected_proposals` and setting `ExecutionStatus`. None of that is reached.

## Observed — existing test coverage cannot detect this

`tests/unit/test_approval_gates.py` contains one test that calls the resume endpoint:

```python
async def test_unauthenticated_list_and_resume_rejected(self, client, mock_storage):
    """The list and resume endpoints also require authentication."""
    ...
    assert resume_resp.status_code == 401
```

The request is rejected at the auth dependency and never reaches `resume_flow`. So resume has coverage proving it is protected, and none proving it works — which is consistent with the failure shipping unnoticed.

## Reproduction

```bash
# 1. enable the gate and restart
#    APPROVAL_GATE_ENABLED=true
#    APPROVAL_REQUIRED_FLOWS=proposal_decision

# 2. create a gated proposal
curl -s -X POST localhost:8001/proposals -H "Content-Type: application/json" \
  -d '{"product_id":"inv-ctv-apex-series","deal_type":"preferreddeal","price":40.0,
       "impressions":1000000,"start_date":"2026-09-01","end_date":"2026-09-30",
       "buyer_id":"lab-buyer-1"}'
# -> status: pending_approval, approval_id: <AID>

# 3. decide it (works)
curl -s -X POST "localhost:8001/approvals/<AID>/decide" \
  -H "Authorization: Bearer <OPERATOR_KEY>" -H "Content-Type: application/json" \
  -d '{"decision":"approve","decided_by":"lab","reason":"repro"}'
# -> 200, decision recorded, approval.granted emitted

# 4. resume it (fails)
curl -s -X POST "localhost:8001/approvals/<AID>/resume" \
  -H "Authorization: Bearer <OPERATOR_KEY>" -w "\nHTTP %{http_code}\n"
# -> Internal Server Error / HTTP 500
```

Repeating step 3 with `"decision":"reject"` on a second approval produces the same 500 at step 4.

## Environment

seller-agent v2.4.2, commit `e5b367d2b780aa4d0e03b6363d744ba3adf2b221`, installed with `uv sync --locked`. CrewAI 1.15.2, resolved from the `>=1.14.4,<2.0.0` constraint. Python 3.12.3, WSL2 Ubuntu 24.04.

## Inferred

**Probable mechanism: CrewAI version drift.** `flow.state` appears to have been assignable in the CrewAI generation this code was written against, and is a read-only property in the lock-resolved 1.15.2. I have not verified which release introduced the change, so this is offered as a likely explanation rather than an established one.

**Not established:** what happens to a decided-but-unresumed flow. The decision persists and the event is emitted, so the audit record is intact. Whether anything downstream treats the proposal as accepted or rejected without the resume step, I have not tested.

## Suggested remedy

Not prescribed — reconstructing or initialising `ProposalHandlingFlow` through CrewAI's supported state mechanism is a maintainer call, and the right approach depends on what that runtime offers in the pinned version.

Worth noting alongside any fix: a test that exercises resume with a valid credential would have caught this. The current coverage asserts only that the endpoint rejects unauthenticated callers.

## Acceptance criteria

- `POST /approvals/{approval_id}/resume` completes for an approved decision and for a rejected decision, returning a structured response rather than an unhandled 500.
- A test exercises the resume path with a valid credential, so that a failure inside `_resume_proposal_flow` is detectable.

aleksUIX:
## Summary

Fixes #67. `POST /approvals/{id}/resume` 500s after a successful decide. `_resume_proposal_flow` assigned `flow.state = ProposalState(**snapshot)`. CrewAI 1.15 made `Flow.state` a read-only `@property`, so `__setattr__` raises `AttributeError: property 'state' of 'ProposalHandlingFlow' object has no setter`. Decide still writes the approval; only resume blows up.

The property reads the `_state` PrivateAttr. Existing tests already hydrate that way (`flow._state = state` in `test_proposal_error_taxonomy.py`). Resume now does the same. In-place mutation of `flow.state.proposal_id` etc. is unchanged.

Empty snapshots use `snapshot or {}` so `ProposalState(**None)` cannot throw.

## Fix

```python
flow = ProposalHandlingFlow()
flow._state = ProposalState(**(snapshot or {}))
```

## Out of scope (deliberate)

- Does not take #65 (`deal_id or ""`, `deal_id_required`) or #69 (rate card).
- Does not change decide, list, or the CrewAI Flow class itself.

## Test plan

- [x] service: resume after approve hydrates snapshot and returns `accepted`
- [x] service: resume after reject returns `rejected` without assigning `flow.state`
- [x] HTTP: authenticated `POST /approvals/{id}/resume` after decide returns 200 (`resumed_from_approval`)
- [x] `uv run --locked ruff check src/` and `ruff format --check src/ tests/` clean

Closes #67


Sirajmx (approved):
> Verified: full unit + integration suite passes, 0 failed, merged onto current main. Live e2e
> confirms it too - resume genuinely 500s on main's CrewAI 1.15 break, 200s clean on this branch.
> Merging.

PR #75, opened 2026-09-14
fix: deals created outside the quote path 500 on read - map internal status to the wire enum (#73)

### LINKED ISSUE #73: GET /api/v1/deals/{deal_id} returns 500 for a deal created through POST /api/v1/deals/from-template
`POST /api/v1/deals/from-template` returns 201 and persists the deal with status `confirmed`. A subsequent `GET /api/v1/deals/{deal_id}` for that same deal returns HTTP 500. The deal is created successfully but cannot be read back through its own endpoint.

`confirmed` is a seller-local status. `DealBookingStatus.CONFIRMED` is declared at `src/ad_seller/models/quotes.py:38`, and the literal is compared against in export, push and active-check logic at `src/ad_seller/services/deal_service.py:827`, `:846` and `:883`. The shared `DealStatus` in `iab-agentic-primitives` v0.5.1 declares twelve members — `proposed`, `negotiating`, `accepted`, `booked`, `active`, `makegood_pending`, `partially_cancelled`, `completed`, `rejected`, `failed`, `cancelled`, `expired` — and `confirmed` is not among them. `internal_deal_to_shared_deal()` passes the persisted value into `DealStatus(...)` with no translation at `src/ad_seller/interfaces/api/contract_mappers.py:319`, raising `ValueError: 'confirmed' is not a valid DealStatus`.

The same route and the same mapper succeed for a deal booked from a quote. That path persists `proposed`, which the published vocabulary contains, and `GET /api/v1/deals/{deal_id}` returns 200.

Observed on v2.4.2 (`e5b367d`) in WSL2 Ubuntu 24.04 and a clean Ubuntu 24.04 container, the container run from a freshly initialised datastore. The script below was run unchanged in both environments.

The script below is the exact file run in both environments:

```text
sha256 209b9112f32e2f16ae5fafdf11ef35954285d3edf091a1474b704204d886f48d
```

It discovers catalogue inventory rather than relying on environment-specific fixture IDs; the two environments have different catalogues (`inv-ctv-apex-series` and `prod-91ddc363`). The status vocabulary is read from the instance's own `/openapi.json` rather than from an installed package, so the comparison is against what that server publishes.

## Reproduction

```bash
#!/usr/bin/env bash
# L reproduction — GET /api/v1/deals/{deal_id} returns 500 for a deal created
# through POST /api/v1/deals/from-template
#
# Published reproduction. Prints what the server does and summarises whether the
# behaviour reproduced. Always exits 0: a build that does NOT reproduce is a valid
# and useful result, not a script failure.
#
# The script establishes the contract-level incompatibility from the wire: the
# from-template response carries status "confirmed", the server's own /openapi.json
# publishes a DealStatus vocabulary that excludes it, and the corresponding read
# returns 500. A bare 500 would not distinguish this from any other failure.
#
# It does not identify the throwing instruction — HTTP alone cannot. The causal
# bridge to the shared-contract mapper's enum coercion sits in the evidence record,
# which carries the source reference and the traceback.
#
# Usage: OPERATOR_KEY=ask_live_... [BASE=host:port] [PRODUCT_ID=xxx] bash repro-l.sh
# Requires: a running seller-agent with a non-empty product catalogue, and an
#   operator API key. The first operator key must be created with
#   `ad-seller create-operator-key` — POST /auth/api-keys/operator itself requires
#   an existing operator credential.
# Creates: one buyer API key, one from-template deal, one quote, one quote-based deal.

H="Content-Type: application/json"
B="${BASE:-localhost:8001}"
RUN="l-$(date +%Y%m%dT%H%M%S)-$$"

# Clear scratch files from any previous run: a failed fetch would otherwise leave
# a stale file in place and let a check pass on last run's data.
rm -f /tmp/l-*.json

[ -n "${OPERATOR_KEY:-}" ] || { echo "SETUP FAILED: set OPERATOR_KEY (see ad-seller create-operator-key)"; exit 0; }

# Discover a product rather than hard-coding a catalogue ID, which is
# environment-specific. Override with PRODUCT_ID if the first product does not
# support a PG deal.
PID="${PRODUCT_ID:-$(curl -fsS "$B/products" | python3 -c '
import json,sys
d=json.load(sys.stdin); p=d.get("products",d)
if not isinstance(p,list) or not p: raise SystemExit("GET /products returned none")
print(p[0]["product_id"])')}"
[ -n "$PID" ] || { echo "SETUP FAILED: no product available"; exit 0; }
echo "product: $PID"

BUYER=$(curl -fsS -X POST "$B/auth/api-keys" -H "$H" -H "X-Api-Key: $OPERATOR_KEY" \
  -d "{\"seat_id\":\"seat-$RUN\",\"seat_name\":\"L repro\"}" \
  | python3 -c '
import json,sys
d=json.load(sys.stdin)
print(next((v for v in d.values() if isinstance(v,str) and v.startswith("ask_")),""))')
[ -n "$BUYER" ] || { echo "SETUP FAILED: could not mint a buyer key (is OPERATOR_KEY valid?)"; exit 0; }
echo "buyer key: minted"

echo
echo "-- 1. create a deal through /api/v1/deals/from-template --"
FT_CODE=$(curl -sS -X POST "$B/api/v1/deals/from-template" -H "$H" -H "X-Api-Key: $BUYER" \
  -d "{\"deal_type\":\"PG\",\"product_id\":\"$PID\",\"impressions\":500000,\"max_cpm\":90.0}" \
  -o /tmp/l-ft.json -w "%{http_code}")
export FT_CODE
echo "HTTP $FT_CODE"
python3 -c "
import json
d=json.load(open('/tmp/l-ft.json'))
print('  deal_id:                      ',d.get('deal_id'))
print('  from-template response status:',repr(d.get('status')))"

FT_ID=$(python3 -c "
import json;print(json.load(open('/tmp/l-ft.json')).get('deal_id') or '')")
[ -n "$FT_ID" ] || { echo "SETUP FAILED: from-template response had no deal_id"; exit 0; }

echo
echo "-- 2. the server's own published DealStatus vocabulary --"
curl -fsS "$B/openapi.json" -o /tmp/l-openapi.json
python3 -c "
import json
d=json.load(open('/tmp/l-openapi.json'))
e=(d.get('components',{}).get('schemas',{}).get('DealStatus') or {}).get('enum') or []
print('  DealStatus enum:',e)
print(\"  contains 'confirmed':\",'confirmed' in e)"

echo
echo "-- 3. read the from-template deal back --"
GET_CODE=$(curl -sS "$B/api/v1/deals/$FT_ID" -o /tmp/l-ft-get.json -w "%{http_code}")
export GET_CODE
echo "HTTP $GET_CODE"
echo "  body: $(head -c 120 /tmp/l-ft-get.json)"

echo
echo "-- 4. control: a quote-based deal through the same route and mapper --"
QT=$(curl -fsS -X POST "$B/api/v1/quotes" -H "$H" \
  -d "{\"idempotency_key\":\"$RUN-q\",\"product_id\":\"$PID\",\"deal_type\":\"PG\",\"impressions\":500000,\"target_cpm\":{\"currency\":\"USD\",\"amount_micros\":20000000}}" \
  | python3 -c "import json,sys;print(json.load(sys.stdin)['quote']['quote_id'])")
[ -n "$QT" ] || { echo "SETUP FAILED: no quote created"; exit 0; }
curl -fsS -X POST "$B/api/v1/deals" -H "$H" \
  -d "{\"idempotency_key\":\"$RUN-d\",\"quote_id\":\"$QT\"}" -o /tmp/l-qb.json
QB_ID=$(python3 -c "
import json;d=json.load(open('/tmp/l-qb.json'));print(d.get('deal',{}).get('deal_id') or '')")
[ -n "$QB_ID" ] || { echo "SETUP FAILED: no quote-based deal created"; exit 0; }
python3 -c "
import json;d=json.load(open('/tmp/l-qb.json'))['deal']
print('  deal_id:                     ',d.get('deal_id'))
print('  quote-based response status: ',repr(d.get('status')))"
QB_CODE=$(curl -sS "$B/api/v1/deals/$QB_ID" -o /tmp/l-qb-get.json -w "%{http_code}")
export QB_CODE
echo "  read-back HTTP $QB_CODE"

echo
echo "-- summary --"
python3 - <<'PYEOF'
import json, os
ft   = json.load(open('/tmp/l-ft.json'))
qb   = json.load(open('/tmp/l-qb.json')).get('deal', {})
oa   = json.load(open('/tmp/l-openapi.json'))
enum = (oa.get('components', {}).get('schemas', {}).get('DealStatus') or {}).get('enum') or []
checks = [
    ("from-template create returned HTTP 201",        os.environ.get('FT_CODE') == '201'),
    ("from-template response status is 'confirmed'",   ft.get('status') == 'confirmed'),
    ("published DealStatus enum is non-empty",        len(enum) > 0),
    ("published DealStatus excludes 'confirmed'",     'confirmed' not in enum),
    ("GET of the from-template deal returned HTTP 500", os.environ.get('GET_CODE') == '500'),
    ("control: quote-based response status is 'proposed'", qb.get('status') == 'proposed'),
    ("control: 'proposed' IS in the published enum",  'proposed' in enum),
    ("control: GET of that deal returned HTTP 200",   os.environ.get('QB_CODE') == '200'),
]
for label, ok in checks:
    print(f"  [{'x' if ok else ' '}] {label}")
print()
print("  REPRODUCED:", "yes" if all(v for _, v in checks) else "NO — see unchecked boxes above")
PYEOF
exit 0
```

## Observed

From-template creation returns HTTP 201 with `status: "confirmed"`. The served `/openapi.json` declares `DealStatus` without `confirmed`. `GET /api/v1/deals/{deal_id}` for that deal returns HTTP 500. The control deal, booked from a quote, persists `proposed`, which is present in the same published enum, and the same route returns HTTP 200.

`POST /api/v1/deals/from-template` requires authentication and returns 401 without a key. The buyer key was minted through `POST /auth/api-keys` using an operator key created with `ad-seller create-operator-key`; the network route `POST /auth/api-keys/operator` itself requires an existing operator credential.

Not tested: `confirmed` is also persisted by `bulk_deal_operations` (`services/deal_service.py:719`), `create_curated_deal` (`:1317`, `:1347`) and `migrate_deal` (`:1410`); only the from-template path was exercised. Three further persisted values are absent from the published `DealStatus` in the same way — `migrated` (`:1445`), `deprecated` (`:1491`) and `registered` (`:1197`) — and whether any of them reaches this mapper was not established.

Closed #57 and #55 concern `create_curated_deal` and `migrate_deal` respectively and both involve the same local literal, but neither covers retrieval through the shared-contract mapper.

atc964:
## Defect

Deals created via the from-template, bulk-create, curated, and migrate paths, and via the MCP create_deal_from_template tool (which calls the same from-template service), are stored with internal status "confirmed" (src/ad_seller/services/deal_service.py:592, 719, 1317, 1410). That value is not a member of the shared iab-agentic-primitives DealStatus wire enum (primitives/lifecycle.py, v0.5.1: proposed, negotiating, accepted, booked, active, makegood_pending, partially_cancelled, completed, rejected, failed, cancelled, expired).

The wire mapper fed the stored string straight into the enum constructor at src/ad_seller/interfaces/api/contract_mappers.py:319 (status=DealStatus(data.get("status", "proposed"))), so GET /api/v1/deals/{deal_id} raised a bare ValueError and returned 500 for every deal created outside the quote-then-book path.

Affected creation paths:

1. POST /api/v1/deals/from-template (deal_service.py:592)
2. POST /api/v1/deals/bulk create operations (deal_service.py:719)
3. POST /api/v1/deals/curated (deal_service.py:1317)
4. The MCP create_deal_from_template tool (interfaces/mcp_server.py:614, same service path)

Two adjacent cases found while fixing, same failure class:

- Deal migration also stores the replacement deal as "confirmed" (deal_service.py:1410), and migrate/deprecate leave the superseded deal as "deprecated" (deal_service.py:1436, 1481), which is likewise absent from the shared enum.
- Bulk-created deals stored no deal_type, product, pricing, or terms at all, so the shared Deal primitive (deal_type is required) could not be built for them even with the status mapped; GET on them failed with a KeyError instead.

## Fix

The mapper is the repo's declared anti-corruption boundary: internal models stay internal, shared types live only at the wire edge, and the two shapes meet in contract_mappers. Statuses are therefore translated on read, and nothing about what deal_service stores changes (rewriting stored statuses would touch every write path and existing rows).

- New INTERNAL_TO_WIRE_STATUS translation in contract_mappers, applied by internal_deal_status_to_wire before the enum constructor: internal "confirmed" reads as "booked" (a confirmed, auto-booked deal is a booked deal on the wire), internal "deprecated" reads as "cancelled" (the closest shared terminal state for a deal buyers must stop targeting).
- Any future internal status with no translation fails loudly with an error naming the offending status and pointing at INTERNAL_TO_WIRE_STATUS, instead of a bare ValueError from the enum constructor.
- Bulk create now carries deal_type, product, pricing, terms, and buyer_tier over from the booked quote so bulk-created deals are readable at all. Bulk deals created before this change still lack deal_type in storage and will surface the loud error on read.

One known gap left out of scope: curated deals created with the schema default deal_type "PMP" (CuratedDealRequest.deal_type, schemas.py:582) still cannot map to the shared DealType enum (PG/PD/PA) on read. That is a deal-type modeling question, not a status one, and deserves its own issue.

## Tests

New tests/unit/test_deal_status_wire_mapping.py covers the mapper translation (confirmed -> booked, deprecated -> cancelled, shared values pass through), the loud unknown-status error, and GET round-trips for deals created via the from-template service path, the bulk endpoint, the curated service path, and a deprecated deal.

- New file: 20 passed
- Related files (deal booking, curated, lineage, mapper ad-formats): 56 passed together
- Full unit suite: 1483 passed, 0 failed
- ruff check and ruff format --check pass on all changed files; the 6 ruff findings in tests/unit/test_pricing_type.py, test_proposal_flow_time_budget.py, and test_quote_validation.py pre-exist on origin/main untouched by this branch

Fixes #73


Sirajmx (dismissed):
> Verified: unit + integration passed (0 failed), merged onto current `main`. Confirms
> the fix is scoped correctly - the mapper (`contract_mappers.py`) is the intended anti-corruption
> boundary between internal storage shapes and the shared wire enum, so translating on read
> (`confirmed`→`booked`, `deprecated`→`cancelled`) rather than rewriting stored data is the right
> shape for this. The curated-deal `deal_type` gap you flagged as out of scope is a real, separate
> issue - agreed it shouldn't block this fix. No conflicts merging onto `main`.

Sirajmx (approved):

PR #77, opened 2026-09-14
fix: booking accepted anonymous callers - require verified buyer matching the quote

atc964:
## Defect

POST /api/v1/deals (routers/deals.py, book_deal handler; deal_service.book_deal at src/ad_seller/services/deal_service.py:192-345) accepted an OPTIONAL API key used only to scope the idempotency key. It never verified the caller against the quote being booked. Anyone who learned a quote_id could book it at its negotiated tier pricing, since booking never checked who the quote was priced for.

Evidence before this fix:

- src/ad_seller/interfaces/api/routers/deals.py:64 (pre-fix): `api_key_record=Depends(deps._get_optional_api_key_record)` — key optional, and nothing downstream rejected an absent key.
- src/ad_seller/services/deal_service.py:192-345 (book_deal): validates quote existence, expiry, and status, but never compares the caller's identity to the quote's recorded buyer.

Every sibling money-moving endpoint (POST /api/v1/quotes, POST /pricing, POST /proposals, POST /proposals/{id}/counter, POST /api/v1/negotiations/messages, POST /api/v1/deals/from-template) requires a verified buyer via `deps._verified_buyer_context` (interfaces/api/deps.py:154). routers/quotes.py:71-87 shows the standard usage pattern this fix mirrors.

## History

Commit 53216a8 (EP-5.2, July 16 2026) added `_verified_buyer_context` verification to every price-moving path. Its own commit message calls out the one deliberate exception: "POST /api/v1/deals books at the already-capped quoted price - no repricing, verified." That rationale covers price integrity (the quoted CPM can't be inflated after the fact) but not caller identity — nothing stopped a different buyer, or an anonymous caller, from presenting (or not presenting) a key and booking someone else's quote.

## Fix

**REST (POST /api/v1/deals):**

- A verified buyer key is now required. Missing key returns 401 `authentication_required`, matching the pattern already used by the sibling `/api/v1/deals/from-template` endpoint in the same file.
- The verified buyer context is built via `deps._verified_buyer_context`, the same helper every other price-moving endpoint uses.
- Ownership is checked against the buyer recorded in `QuoteHistoryStore` at quote-creation time (this store already records `buyer_id` per quote_id for a different purpose — CPM-tolerance verification — and is the only durable record of who a quote was priced for). A valid key belonging to a different buyer than the quote's recorded owner is rejected with 403 `buyer_mismatch`.
- Tier consistency is also enforced independently: the caller's verified tier must be at least as high as the tier the quote was priced at (403 `buyer_tier_mismatch`). This is defense in depth for quotes with no history record on file (for example a quote seeded outside the normal create-quote path).
- A quote priced at the PUBLIC tier asserted no identity at quote-creation time, so there is nothing buyer-specific to steal — any authenticated buyer may still book it. Only fully anonymous callers are rejected for public-tier quotes.
- A nonexistent quote_id still returns 404 unchanged (checked before ownership, since there's nothing to check ownership against).
- Idempotency scoping (FD-12) is unaffected — the key is still scoped by the buyer's identity, which is now guaranteed present.

**MCP twin (create_deal_from_template, interfaces/mcp_server.py ~613-649):**

This tool booked with no auth at all — it built a `_public_context()` in-process rather than reading credentials from the request, even though its REST twin (`/api/v1/deals/from-template`) already required a key. It is now gated with `_deny_unless_operator` (mcp_server.py:136), consistent with the rest of the MCP server's privileged write tools (inventory sync, operator key management, order transitions, etc).

## Breaking / behavior change

Integrators must now present a buyer API key when calling POST /api/v1/deals. Previously anonymous callers could book any quote at its quoted price; this is no longer possible. A key that does not belong to the buyer the quote was priced for is also rejected, even if it is a valid, active key.

The MCP `create_deal_from_template` tool now requires an operator-role key over HTTP transports (stdio/local process access is unaffected, matching every other admin MCP tool). This is a stricter gate than the REST endpoint's buyer-key requirement — deliberate, since the MCP surface is treated as the seller's own operational/admin console throughout this server, not a buyer-facing surface.

## Tests

Added/updated across four files:

- tests/unit/test_deal_booking_endpoints.py — new `TestBookDealRequiresVerifiedBuyer` class: anonymous booking rejected (401), valid key but wrong buyer rejected (403 buyer_mismatch), lower-tier buyer rejected (403 buyer_tier_mismatch), owning buyer books successfully, public-tier quote bookable by any authenticated buyer. 11 pre-existing tests in `TestBookDeal`/`TestQuoteToDealFlow` updated to authenticate as the quote's buyer (they previously booked anonymously); one of those (`test_idempotency_key_scoped_per_buyer`) also had its mock buyer identity corrected to a full seat/agency/advertiser chain so it resolves to ADVERTISER tier under the new tier check.
- tests/unit/test_deal_booking_snapshot.py — `client`/`mock_storage` fixtures updated: an authenticated advertiser-tier key by default (8 tests), and the mock storage now wires generic get/set (needed by the new ownership check's QuoteHistoryStore lookup).
- tests/unit/test_audience_plan_validation.py — `http_client` fixture updated to present an authenticated key (2 tests; the quote there is public-tier, so any authenticated key satisfies both the auth gate and tier check).
- tests/unit/test_operator_auth.py — new `TestMcpCreateDealFromTemplateGated` class mirroring the existing `TestMcpTransitionOrderGated` pattern: denied without a key (service untouched), denied with a buyer-role key, allowed with an operator-role key.

15 existing tests were updated to authenticate (11 in test_deal_booking_endpoints.py needed a new override added, 1 had its buyer identity corrected, 8 in test_deal_booking_snapshot.py via the shared fixture, 2 in test_audience_plan_validation.py via the shared fixture — some tests share fixtures so the file-level fixture changes cover multiple tests each).

## Verification

```
uv run ruff check <changed files>       -> All checks passed!
uv run ruff format --check <changed files> -> 6 files already formatted
uv run pytest tests/unit/ -q --timeout=30 --timeout-method=thread
  -> 1471 passed
```

Full-repo `ruff check .` / `ruff format --check .` show pre-existing, unrelated drift on `origin/main` (tests/unit/test_quote_validation.py, examples/rivian_r2_linear_tv.py, patches/*.py) — confirmed via `git diff --stat origin/main` that none of those files are touched by this branch.

## CHANGELOG

Added under [Unreleased] / Changed, matching existing formatting.


Sirajmx (approved):
> Verified: unit + integration passed (0 failed), merged onto current `main`. This is a
> real fix, not a hardening exercise - confirmed the defect first: booking accepted a missing or
> wrong-buyer key on `POST /api/v1/deals`, letting anyone who learned a `quote_id` book someone
> else's negotiated price. The fix requires a verified buyer key matching the quote's recorded
> owner, with a public-tier quote (no identity asserted at quote time) still bookable by any
> authenticated buyer, and gates the MCP `create_deal_from_template` twin the same way `deals.py`
> already required its REST counterpart. No conflicts merging onto `main`.

PR #78, opened 2026-09-14
feat: wire the operator rate card into pricing (issue #69)

### LINKED ISSUE #69: Rate card is writable but is not read by the quote, deal-booking or negotiation pricing paths

atc964:
## Summary

Closes #69. The operator rate card (`PUT /api/v1/rate-card`, and the MCP `update_rate_card` tool) has always been writable, but nothing in pricing ever read it: quotes, from-template bookings, and negotiation all priced off the catalog product's `base_cpm`/`floor_cpm` directly. This PR wires the stored rate card into every pricing surface.

**This is a disclosed behavior change**, per the plan agreed on #69: once an operator sets a rate card entry, it now actually affects quoted/booked/negotiated prices for matching products, where before it was silently ignored.

## Design

A stored rate card entry whose `inventory_type` matches a product overrides that product's catalog `base_cpm` as the starting price for:

- quoting (`POST /api/v1/quotes`, `POST /pricing`)
- from-template booking (`POST /api/v1/deals/from-template`)
- negotiation anchoring (proposal evaluation/counter and cold-start `POST /api/v1/negotiations/messages`)

Rules:

- **Floor still applies.** An override below the product's `floor_cpm` is clamped up to the floor — the rate card can raise or lower the effective starting price, but can never price below floor.
- **No matching entry → catalog pricing, unchanged.** A product whose `inventory_type` has no rate card entry prices exactly as it does today.
- **No stored rate card at all → identical to today.** Byte-for-byte the pre-#69 behavior.
- **Unpriced products stay unpriced.** A product with neither `base_cpm` nor `floor_cpm` (the honest-availability 422 path) is not put on sale by a rate card entry — the rate card overrides an *existing* catalog price, it doesn't manufacture one for a product the catalog declares has no price.
- **Provenance logging.** An applied override (and a floor-clamp) is logged at `info` level with the product id, inventory type, catalog price, and override price.

### Implementation

All of the above lives in one new module, `src/ad_seller/services/rate_card_service.py`, consulted by base-price resolution instead of duplicating the lookup at each call site:

- `resolve_base_cpm(product)` — used by `quote_service.get_pricing`, `quote_service.create_quote`, and `deal_service.create_deal_from_template`. Wraps `catalog_service.priceable_cpm` (so the 422-unpriceable gate is unchanged) and layers the override/clamp on top.
- `resolve_negotiation_anchor(product_data)` — used by `negotiation_service.counter_proposal` (cold-start / quote-led negotiation) and `ProposalHandlingFlow.evaluate_pricing` (the primary anchor for a freshly-submitted proposal, which `ProposalHandlingFlow.generate_counter_terms` then reuses via `state.evaluation.recommended_price` rather than re-resolving). Same match/clamp logic, adapted to the dict-shaped product data negotiation already works with, and keeping negotiation's existing default-to-0 semantics for missing pricing rather than introducing a new 422 there.

**Match key:** `RateCardEntry.inventory_type` — the only match key the model actually supports (`interfaces/api/schemas.RateCardEntry` has `inventory_type`/`base_cpm`/`currency`/`effective_date`/`notes`; no `product_id` or tier field). This is effectively "media type" from the design brief.

### Read-side fix

`GET /api/v1/rate-card` used to return the same response shape whether or not an operator had ever set a card — the unset-state response (generic reference defaults) was indistinguishable from a real stored card except for an easy-to-miss `updated_at: "default"` sentinel. `RateCardResponse` now carries an explicit `source: "stored" | "defaults"` field (`updated_at` is `None` for the unset case), and the MCP `get_rate_card` tool got the same treatment for consistency.

## Design deviations from the brief

- **Deliberately-unpriced products stay 422, even with a matching rate card entry.** The brief didn't specify this case explicitly; I chose to keep `catalog_service.priceable_cpm`'s honest-availability 422 as the gate a rate card entry can't bypass, since "never a fabricated price" is a repeated, load-bearing invariant elsewhere in this codebase (avails, quotes, curated deals). Covered by `TestResolveBaseCpm::test_unpriced_product_still_422s_even_with_matching_entry`.
- **`deal_service.create_curated_deal` (curator overlay pricing) is untouched.** It's a fourth `catalog_service.priceable_cpm` call site the brief didn't mention (only quoting, from-template booking, and negotiation were in scope). Flagging it here in case a maintainer wants the rate card to reach curated deals too — the resolver is a two-line change away (`rate_card_service.resolve_base_cpm(product)` in place of `catalog_service.priceable_cpm(product)` at `deal_service.py`'s `create_curated_deal`).
- **`GetRateCardTool` (AgentCore) left unchanged, as instructed if not trivial.** It recomputes a rate-card-shaped view from `GET /products` rather than reading the stored rate card at all, so it never reflects an override today. Aligning it isn't a trivial resolver swap — it's a separate HTTP tool with no access to the resolver in-process, and reflecting overrides through it would mean either a second HTTP call to `GET /api/v1/rate-card` and merging client-side, or changing what `/products` serializes (out of scope here). Left as-is.
- **`quote_service.get_pricing` became `async`**, since resolving a rate card override means an awaited storage read. Updated its three call sites (`POST /pricing`, the MCP `get_pricing` tool, and the `ad-seller price` CLI command, which now runs it via `asyncio.run` following the same pattern as this file's other storage-touching commands).

## Test plan

New `tests/unit/test_rate_card_pricing.py` (21 tests):

- `TestResolveBaseCpm` — the resolver itself: override applied, override clamped to floor, no matching entry falls back to catalog, no stored rate card falls back to catalog (regression pair), unpriced product still 422s even with a matching entry.
- `TestQuoteUsesRateCard` — same cases through `quote_service.create_quote` and `quote_service.get_pricing`.
- `TestFromTemplateBookingUsesRateCard` — same cases through `deal_service.create_deal_from_template`.
- `TestNegotiationAnchorReflectsOverride` — cold-start `counter_proposal` anchor (override, floor-clamp, no-card regression pair) and `ProposalHandlingFlow.evaluate_pricing`'s `recommended_price`.
- `TestIdempotentReplayUnaffected` — a rate-card-priced quote books and replays idempotently through `POST /api/v1/deals` exactly like any other quote (FD-12 untouched).
- `TestRateCardReadSide` — `GET`/`PUT /api/v1/rate-card` `source` field (`defaults` vs `stored`).

Also fixed one existing test (`test_pricing_for_unpriced_default_product_is_422` in `tests/unit/test_default_catalog_enrichment.py`) that called `quote_service.get_pricing` synchronously — updated to `await` it with storage mocked, since the resolver now touches storage.

**Full suite:** `1484 passed` (`tests/unit`, `uv run pytest tests/unit --timeout=600`). `ruff check` and `ruff format --check` clean on all changed files.

## Changelog

Added under `[Unreleased] / Changed` in `CHANGELOG.md`.


Sirajmx (commented):
## PR #78 - Heads up (will become blocking once #77 merges)

> Not an issue with #78 on its own — it's green today. Flagging ahead of time since #77 is merging
> first: merges cleanly with #77 at the code level (only shared file is `mcp_server.py`, different
> regions, no textual conflict), but once #77 is in and this rebases onto it,
> `TestIdempotentReplayUnaffected::test_booking_replay_of_a_rate_card_priced_quote` will start
> failing — it books unauthenticated, and #77 requires a verified buyer key on `POST /api/v1/deals`.
> Confirmed this locally by merging the two together: the test gets 401 instead of the 200 it
> expects.
>
> Direction: authenticate the client for this test's two booking calls, same pattern as #77's own
> `test_public_tier_quote_bookable_by_any_authenticated_buyer` (the quote here is public-tier via
> `_public_context()`, so any authenticated buyer works):
>
> ```python
> from unittest.mock import MagicMock
> from ad_seller.models.buyer_identity import BuyerIdentity
>
> app.dependency_overrides[_get_optional_api_key_record] = lambda: MagicMock(identity=BuyerIdentity())
> ```
>
> right before the two `client.post("/api/v1/deals", ...)` calls. Verified this fix locally —
> full suite green ( unit + integration passed, 0 failed) with #78 merged onto #77+#75+#80.
> No other interaction found across the five open PRs. Nothing else blocking.

Sirajmx (commented):
> Adversarial review found a real gap in the floor guarantee. The PR states "the rate card can
> raise or lower the effective starting price, but can never price below floor" - true only when
> the product has `floor_cpm` set. `ProductDefinition.floor_cpm` is `Optional[float] = None`, and
> `RateCardEntry.base_cpm` has no positivity constraint anywhere (not on the model, not on the
> operator-gated `PUT /api/v1/rate-card` endpoint). Reproduced live: a product with
> `base_cpm=35.0, floor_cpm=None` + a rate card entry `base_cpm=-5.0` → `resolve_base_cpm` returns
> `-5.0` directly (the clamp in `_apply_override` no-ops when `floor_cpm is None`).
> Quotes/bookings are incidentally saved by `PricingRulesEngine`'s unrelated `global_floor_cpm=1.0`
> default — but negotiation has no such backstop: I ran `negotiation_service.counter_proposal`
> end-to-end with the same adversarial state and the stored `NegotiationHistory` came back with
> `base_price=-5.0, floor_price=0.0` — a genuinely negative anchor with zero protection anywhere
> in that call path.
>
> Only exploitable by a trusted operator (the PUT is `_require_operator_api_key_record`-gated),
> but a typo or a compromised operator key now propagates an unbounded negative price straight
> into negotiation — the exact thing the "never a fabricated price" invariant elsewhere in this
> codebase exists to prevent.
>
> Direction: `_apply_override` should also floor at 0 (or `PricingRulesEngine`'s
> `global_floor_cpm`) when the product's own `floor_cpm` is `None`, not skip clamping entirely;
> and/or add a `gt=0` constraint to `RateCardEntry.base_cpm` so this can't be stored in the first
> place.
>
> Everything else here checks out — verified 1517 unit + 39 integration passed (0 failed) on the
> latest rebase onto current `main`, no conflicts anywhere. Requesting changes on the floor-safety
> gap above before this merges; happy to re-review once that's in.

Sirajmx (approved):
> The floor-safety gap from adversarial review is fixed (`074b88a`) — verified both closure
> points live. `RateCardEntry.base_cpm` now has `Field(gt=0)`; tried
> constructing one with `-5.0` directly and got a genuine `ValidationError`, can no longer be
> stored at all. Separately, `_apply_override` now resolves an effective floor via
> `TieredPricingConfig.global_floor_cpm` whenever the product has no `floor_cpm` of its own,
> instead of skipping the clamp; re-ran the exact adversarial repro (product with `base_cpm=35.0,
> floor_cpm=None` + a rate card entry already at `-5.0`) through `negotiation_service.counter_proposal`
> end-to-end and the stored anchor now comes back at `base_price=1.0` (the global floor), not
> `-5.0`. Both of the author's own new regression tests pass too.
>
> Full suite green: unit + integration passed (0 failed) on the latest rebase onto current
> `main`, no conflicts anywhere.

PR #80, opened 2026-09-15
fix: add --quiet to create-operator-key for scripting

### LINKED ISSUE #79: CLI: add --quiet to create-operator-key, matching the buyer agent
## Summary

`ad-seller create-operator-key` has no `--quiet` flag, so there is no supported way to capture the minted key from a script. Its counterpart in `buyer-agent` does have one. Adding it would make the two CLIs symmetric and remove the need for callers to parse console output.

## Current behaviour

`src/ad_seller/interfaces/cli/main.py` -> `create-operator-key` accepts `--label` and `--expires-in-days`, then renders the result through a rich `Panel` plus several `console.print` lines, with the key embedded in that formatted output. There is no machine-readable mode.

The practical consequence is that any automated bootstrap (a container entrypoint, a provisioning script, a first-run setup step) has to scrape the key out of the rendered text, typically by matching the documented `ask_live_` prefix from `models/api_key.py`. That works, but it couples callers to the CLI's presentation layer: a cosmetic change to the panel, the wrapping width, or the surrounding copy silently breaks credential bootstrap, and the failure shows up later as an authentication error rather than as an obvious CLI problem.

This matters more since operator keys became mandatory for the admin surface, because minting one is now a required setup step rather than an optional convenience.

## Suggested fix

Mirror the buyer agent, which already solves this. In `buyer-agent`, `ad-buyer create-operator-key` takes a `--quiet` / `-q` flag documented as \"Print only the key, for scripting\", and when set it prints the bare key to stdout and returns before rendering any panel.

The equivalent here would be a `--quiet` / `-q` option that prints only \`response.api_key\` and returns early, leaving the existing rich output as the default. Worth flushing stdout explicitly, as the buyer implementation does, so the value survives command substitution.

That would let callers do:

\`\`\`bash
KEY=\$(ad-seller create-operator-key --label ci --quiet)
\`\`\`

## Scope

Small and self-contained: one new option on one command, no change to the default output, no change to key generation or storage. Probably suitable as a good first issue.

aleksUIX:
## Summary

Fixes #79. `ad-seller create-operator-key` had no machine-readable mode, so bootstrap scripts had to scrape the minted key out of the rich panel. `ad-buyer create-operator-key` already has `--quiet` / `-q`.

This adds the same flag: print only `response.api_key` and return before the panel. Stdout is flushed so the value survives command substitution.

```bash
KEY=$(ad-seller create-operator-key --label ci --quiet)
```

Default interactive output is unchanged.

## Test plan

- [x] `uv run --locked pytest tests/unit/test_operator_auth.py::TestCliBootstrap -q` (7 passed)
- [x] `uv run --locked ruff check` / `ruff format` on the touched files


Sirajmx (commented):
> Verified: unit + integration passed (0 failed), merged onto current `main`. Small,
> exactly matches the buyer-agent's existing `--quiet` pattern, default interactive output
> unchanged. No conflicts merging onto `main`.

atc964 (approved):
Approved. Thanks for turning this around so quickly.

Verified the implementation matches the buyer agent's `--quiet` behaviour: bare key
to stdout, `flush=True` so it survives command substitution, early return before the
panel, and the default interactive output untouched. The test asserting stdout is
exactly the key and stderr is empty is the right assertion for this.

Two non-blocking nits, neither worth holding the merge for:

1. `if quiet is True:` can just be `if quiet:`.
2. In `docs/api/authentication.md` the two commands now sit in one block and read as a
   sequence to run in order, which would mint two separate operator keys. Splitting
   them, or showing the `--quiet` form as an alternative rather than a follow-on, would
   read more clearly.

Happy for this to merge as-is.

PR #90, opened 2026-09-16
fix: single-source the version so served surfaces cannot report a stale one

atc964:
## The defect

`ad_seller.__version__` is meant to be the single source of truth for the version, but two routers each repeated the literal `"2.4.2"`, and both serve it to callers:

| Surface | File | Who reads it |
| --- | --- | --- |
| `GET /` | `interfaces/api/routers/admin.py` | API clients, humans poking the root |
| `GET /.well-known/agent.json` | `interfaces/api/routers/registry.py` | registries and buyer agents |

The agent card is the one that matters most: registries and buyer agents fetch it to decide what this seller is and what it supports, so a stale version there misinforms every counterparty rather than just a curious caller.

This was invisible in CI. Nothing asserted the served version against the package version, so the package could install as a new version while the running endpoints kept reporting the old one, and bug reports would name a version nobody is running. It already survived two bumps unnoticed (v2.4.0 to v2.4.1 to v2.4.2) and was caught while scoping the next one, which would have shipped a seller advertising 2.4.2 to every caller.

## The fix

Both routers now import `__version__` from the package and pass it at the call site. Four lines.

**The version itself is unchanged at 2.4.2.** The release is not cut yet, so nothing here bumps `pyproject.toml`, `src/ad_seller/__init__.py`, `uv.lock`, or any CHANGELOG heading. This PR only removes the duplication so that whoever does cut the release gets the bump propagated to the served surfaces for free.

## The guard

New `tests/unit/test_version_single_source.py`:

1. **Scan test** walks every `.py` under `src/ad_seller` and fails if the *current* `__version__` value appears anywhere except the top-level `__init__.py`, naming the offending file and line. It matches the exact current value rather than a semver pattern, so unrelated numbers (OpenDirect spec dialects, protocol versions, dependency pins) cannot trip it.
2. **Call-site test** asserts both routers import `__version__` and actually use it below the router definition, catching the case where the import stays but a literal creeps back in.

Negative control: putting `version="2.4.2"` back into `registry.py` makes the scan test fail with

```
interfaces/api/routers/registry.py:52: version="2.4.2",
```

so the guard genuinely fails when the defect is present. Control was reverted.

## Verification

- Full unit suite: 1521 passed. The existing agent-card tests (`test_endpoint_no_flow_kickoff.py`, `test_capability_audience_block.py`) cover `/.well-known/agent.json` and still pass, so the new router import causes no circular-import or import-order problem with the `_telemetry_shim` must-be-first import.
- Both endpoints checked live over ASGI: `GET /` and `GET /.well-known/agent.json` each return 200 and report `ad_seller.__version__`.
- `ruff check src/` and `ruff format --check src/ tests/` clean.
- No existing test needed changing; none of them asserted the hardcoded literal.

Bead: seller-nad


Kishore-2572 (commented):
src/ad_seller/__init__.py
src/ad_seller/interfaces/api/main.py

@atc964 This dynamic versioning also need to updated in above two files as well. 
I will be updating that as part of release workflow implementation in branch `add-release-versioning`

PR #91, opened 2026-09-16
fix: treat target_cpm as advisory so it cannot move the quoted price

atc964:
## The defect

`quote_service.create_quote` overwrote its own computed `final_cpm` with `request.target_cpm` whenever that value cleared the floors:

```python
final_cpm = decision.final_price
if request.target_cpm is not None:
    acceptable, _ = engine.is_price_acceptable(...)
    if acceptable:
        final_cpm = request.target_cpm
```

This was wrong in **both** directions.

**Above the seller's price it overcharged**, and did so while the rest of the pricing envelope went on advertising the discount that had just been discarded. A real booking quoted `"Advertiser tier: -15% | Final price: $12.75 CPM"` with `tier_discount_pct: 15.0`, and billed **$15.00**. A buyer whose own rate card led it to send a conservative `target_cpm` of $15.00 was charged its own estimate rather than the seller's better price, and the response it received contradicted the number it was billed.

**Below the seller's price it conceded margin automatically.** Any buyer naming a number above the floor got that number, because `is_price_acceptable` is a floor check, not a pricing decision. A $11.00 target against a $12.75 price simply became $11.00, with no negotiation and no record of a concession.

## The fix

The seller's computed price is always the quoted price. `final_cpm = decision.final_price`, unconditionally. A target below the quoted price is logged at INFO as a negotiation signal and otherwise ignored.

This aligns behaviour with the contract: `QuoteRequest.target_cpm` is defined in the shared protocol spec as *"Buyer's desired CPM ... Advisory"*. It is the buyer's statement of what it expects to pay, formed from its own rate card and spend history before the seller has priced anything. It is an input to whether we quote or negotiate, never a figure that sets the price. A discount is a decision the seller makes, not a number the buyer can assert.

If we ever want to honour a lower target automatically, that is a deliberate pricing policy that belongs in `PricingRulesEngine`, where it would be reflected in the rationale, rather than applied silently at the service layer.

## An existing test asserted the old behaviour and has been rewritten

Reviewers should not be surprised by this. `tests/unit/test_quote_endpoints.py::TestCreateQuote::test_target_cpm_accepted_when_above_floor` asserted that a $32.00 `target_cpm` against this product's $35.00 price came back as **$32.00**, because the old code accepted any buyer number clearing the floors. Its name ("accepted") shows the mental model: a buyer price was treated as an offer to accept. That conceded margin on request and bypassed negotiation.

It is now `test_target_cpm_below_our_price_does_not_lower_the_quote` and expects $35.00. A companion `test_target_cpm_above_our_price_does_not_raise_the_quote` covers the overcharge direction with a $40.00 target, also expecting $35.00.

No other test in the suite depended on the old behaviour. The full unit suite is green at 1537 passed.

## New coverage

`tests/unit/test_target_cpm_advisory.py` exercises the matrix at service level using the shape of the real incident (base $15.00, product floor $10.00, advertiser tier, so the seller's price is $12.75): no target, target above, below, equal to, and beneath the floor, plus public tier. It also adds **pricing-envelope consistency** tests that parse the dollar figure out of `rationale` and compare it to `final_cpm`, and that check `base_cpm` reduced by the advertised tier and volume discounts equals `final_cpm`. Those are the assertions that would have caught the original defect from the response alone, since it was visible there: a rationale stating $12.75 sitting next to a `final_cpm` of $15.00. A booked/stored-quote test confirms the persisted quote carries the quoted price.

The fix was verified with a negative control: restoring the old override makes 11 of these tests fail, including both endpoint tests and the rationale-consistency parametrisations.

## Note: `PricingRulesEngine.is_price_acceptable` now has no caller in `src/`

Deleting it would be wrong, so this PR does not. It is donated public reference code that external implementers may import, and it keeps direct unit coverage in `tests/unit/test_engines.py`. **Flagging it for a later deprecation decision** rather than making that call here.

## Out of scope for this PR: the product floor is not enforced in `calculate_price`

Noted while working in this code, **not fixed here**, to keep this PR to one decision.

`PricingRulesEngine.calculate_price` clamps only to `self._config.global_floor_cpm` (default $1.00) and never receives or applies the product's own `floor_cpm`, while the rate card clamp and `is_price_acceptable` both use the product floor. Confirmed against the real code: a product with `base_cpm` $15.00 and `floor_cpm` $14.00, quoted to an advertiser-tier buyer, yields **$12.75, below its own floor**, end to end through `create_quote`. The engine will then reject its own quoted price if you hand it back: `is_price_acceptable($12.75, product_floor=$14.00)` returns `False, "Below product floor ($14.0 CPM)"`.

Worth a separate issue.


atc964:
**Relaying a review from Siraj, who reviewed in full and verified locally but is offline** (posting so it is on the record here rather than only in chat). His words, condensed:

> Confirmed the spec claim directly: `iab-agentic-primitives` genuinely defines `target_cpm` as "Advisory." Reproduced with a negative control myself; reverting to the old override behaviour fails exactly 11 tests, matching the stated count. The rationale/final_cpm consistency tests are the right addition, they'd have caught the original defect from the response alone.
>
> One real, verified concern before this merges: buyer-agent's negotiate-above-ceiling flow (`MultiSellerOrchestrator._negotiate_quote`) depends on the exact behaviour this PR removes. After negotiating to an agreed price it re-quotes with `target_cpm=<agreed price>`; its own comment calls that "the only contract-clean way to book a negotiated price today." Post-#91 that re-quote returns the seller's standard price, and the buyer only checks the result is `<= max_cpm`, not that it equals the negotiated price. So a real negotiation silently succeeds at the wrong price, logged as "Negotiation succeeded," with no error anywhere. Same blind spot as the #77/#132 situation, and it needs a matching buyer-side fix before #91 ships.

**Confirmed, and it is worse than the review found.**

The re-quote and the ceiling-only check are exactly as described: `multi_seller.py` sends `target_cpm=agreed`, then the only guard is `final_cpm is None or final_cpm > max_cpm`. It never compares against `agreed`.

The seller does appear to have a clean alternative. `deal_service.py:246` honours an ACCEPTED negotiation on the booked quote and strikes the agreed price, and its comment even calls the buyer's re-quote "the buyer's historical workaround." **That block is unreachable.** Negotiations are stored keyed by proposal id (`negotiation_service.py:122/270/352`, and `storage/base.py:278` declares the parameter as `proposal_id`), while the booking looks up by `quote_id`. The id spaces cannot overlap: `prop-{uuid4 hex[:8]}` versus `qt-{uuid4 hex[:12]}`. So the lookup always returns `None`, and a booked deal has never carried a negotiated price.

Which means the buyer's `target_cpm` re-quote is not a workaround sitting beside a clean path. It is currently the **only** mechanism that books a negotiated price, and this PR removes it.

**So this PR is on hold**, not because the change is wrong (it matches the spec, and the review confirmed the reasoning independently) but because the replacement it assumes does not exist yet. Filed separately on the seller side. Three candidate shapes: index negotiations by the quote they concern, carry a negotiation reference on the booking request, or write the accepted price onto the quote when the negotiation concludes so booking needs no lookup. Whichever is chosen, the booked price must be asserted to equal the agreed price rather than merely sit under the buyer's ceiling.

Also worth recording: the rig gate passed yesterday without catching any of this, because it runs S1, which books without negotiating. S2 is the negotiation scenario and has never been run against these changes. That gap is the actionable lesson, more than the bug itself.

atc964:
**Correction to my earlier comment**, narrowing a claim I overstated.

I said the seller's honour-the-agreed-price block was unreachable dead code. That is too strong. `routers/negotiation.py:139` derives the storage key as `message.proposal_id or message.negotiation_id or message.quote_id`, so a **quote-led** negotiation is stored under the quote id, and `deal_service.py:246`'s lookup by `quote_id` finds it. The comment just above that line says so explicitly. The seller's design works for quote-led negotiations.

The accurate claim is narrower: **it never fires for our buyer**, which is proposal-led. `negotiation/client.py:191, 245, 294` always send `proposal_id` (plus `negotiation_id` on continuation) and never `quote_id`, so the negotiation is keyed by proposal and the booking lookup misses every time. The practical consequence for this PR is unchanged: the `target_cpm` re-quote is the only thing currently delivering a negotiated price to a booking, and this PR removes it.

One further finding while verifying, same root cause and worth fixing alongside: the storage key comes from whichever id a given message happens to carry, so two messages in one negotiation can resolve to different keys. A continuation carrying only `negotiation_id` would key off the `neg-` id and miss the round-one record, silently restarting the negotiation. Our buyer avoids this only because it sends `proposal_id` and `negotiation_id` together, so `proposal_id` keeps winning. That is luck rather than design.

PR #92, opened 2026-09-16
fix: never discount a quote below the product's own floor

atc964:
## The bug

`PricingRulesEngine.calculate_price` clamped the discounted price to
`self._config.global_floor_cpm` ($1.00 by default) and was **never passed the
product's own `floor_cpm`** at all. So tier and volume discounts could quote
below a floor the product explicitly declares.

Measured on the real engine:

| | |
|---|---|
| base CPM | **$15.00** |
| product `floor_cpm` | **$14.00** |
| ADVERTISER tier (-15%) | **$12.75** quoted |

$12.75 is $1.25 under a floor the product declares.

## The self-contradiction

The engine then **rejects its own quoted price**. `is_price_acceptable` *does*
check the product floor:

```
is_price_acceptable(offered_price=12.75, product_floor=14.0)
-> (False, "Below product floor ($14.0 CPM)")
```

`calculate_price` and `is_price_acceptable` are two halves of the same engine,
so a price the first produces must be one the second accepts or the seller
contradicts itself. The discount waterfall was the outlier here: two
neighbouring paths already respected the product floor (`rate_card_service`
clamps a rate-card override up to it, `is_price_acceptable` checks it).

## The fix

The effective floor is now the **higher** of the global backstop and the
product's own floor:

```python
floor = self._config.global_floor_cpm
if product_floor is not None:
    floor = max(floor, product_floor)
```

The clamp went **inside the engine**, with `product_floor` as an optional
parameter, so all three call sites inherit it rather than each repeating the
same check:

- `quote_service.get_pricing`
- `quote_service.create_quote`
- `deal_service.create_deal_from_template`

**Products with no floor are unaffected.** `product_floor` is optional and
defaults to `None`; when it is absent the global floor remains the only
backstop, exactly as before. Nothing changes for a product whose floor sits
below the discounted price either, since the floor is a minimum and not a
target.

## Tests

`tests/unit/test_product_floor_clamp.py` covers engine-level clamping, the
no-over-clamping case, the `None`-floor fallback, higher-of-the-two, two
service-level tests through `create_quote`, and a parametrized invariant that
the engine never quotes a price its own `is_price_acceptable` would reject.

Verified by negative control: reverting the clamp to the global floor alone
fails 4 of them, including the invariant, with

```
AssertionError: engine quoted $12.75 against a $20.0 product floor
and then rejected it: Below product floor ($20.0 CPM)
```

Full unit suite: **1548 passed**. No existing test encoded the old behaviour,
so no assertion was changed or weakened. `ruff check src/` and
`ruff format --check src/ tests/` both clean.

## Stacking

> **This PR is stacked on #91** (`fix/target-cpm-advisory`) and **should merge
> after it.** The base is set to `fix/target-cpm-advisory` so the diff here
> shows only the floor change; GitHub retargets this to `main` automatically
> once #91 merges. #91 edits `quote_service.py` a few lines below one of these
> hunks, and the rebase applied with no conflict.


atc964:
Added tests pinning one side effect of this change, so a reviewer can see it was deliberate rather than incidental.

The clamp sits after the rule-override branch in `calculate_price`, which means a `PricingRule.base_price_override` below the product floor is now clamped up to the floor. That was previously untested in either direction, so `TestRuleOverridesAreAlsoFloored` now covers both:

- **override below the floor is clamped up**: override `8.0` against a `14.0` product floor yields `14.0`, and both `Price override` and `Floor enforced` appear in `applied_rules`
- **override above the floor is untouched**: override `22.0` against the same `14.0` floor stays `22.0`

This is consistent with `rate_card_service`, which already clamps a rate card override up to the product floor: a floor is a floor, and an override is not an exemption from it.

Pinning it rather than leaving it implicit for the same reason this PR exists at all. An untested side effect is how the `target_cpm` defect survived, and it reads as accidental to whoever touches this next. If the intent ever changes to "an explicit operator override wins over the floor", these two tests are what has to change with it.

Tests only, no `src/` change. Full unit suite green (1550 passed), `ruff check src/` and `ruff format --check src/ tests/` both clean.

PR #94, opened 2026-09-17
test: give the suite a per-run temporary database

atc964:
## The defect

The test suite resolved its storage path from `DATABASE_URL` / a local `.env`, so it wrote a SQLite file into the working tree and then reused it on the next run. State written by run N was visible to run N+1, which made the **first** run green and **every later run fail** in the same working tree.

Concretely, on pristine `origin/main`:

```
$ pytest tests/unit/test_trust_tier_verification.py -q
24 passed in 1.52s
$ ls *.db
ad_seller_smoke.db
$ pytest tests/unit/test_trust_tier_verification.py -q
>       ctx = counter.await_args.kwargs["buyer_context"]
E       AttributeError: 'NoneType' object has no attribute 'kwargs'
1 failed, 23 passed
```

`TestNegotiationMessagesCeiling::test_self_asserted_advertiser_identity_is_floored` posts `idempotency_key: "idem-1"`. On the rerun the negotiation idempotency short-circuit finds that key already persisted, replays the cached 200, and never calls the mocked `counter_proposal` — so `await_args` is `None`. Deleting the database restores a full green run.

CI never sees this because every job starts from a clean checkout, which is exactly why it survived. It predates all current branches and has already cost debugging time twice, because the failure looks like it was caused by whatever branch you happen to be on.

## The fix

Point the whole run at a fresh temporary directory in the root `conftest.py`, and remove that directory in `pytest_unconfigure`.

Isolation is now **structural** rather than per-test: no test has to remember to pick a unique idempotency key, and nothing is left in the working tree to leak into the next run. The narrower alternative — making the idempotency-sensitive tests use unique keys — was rejected because it leaves the underlying isolation defect in place for the next test that trips it.

Two implementation notes, both explained in comments at the call sites:

- The override is **module-level**, before the first `ad_seller` import, because `get_settings` is an `lru_cache` singleton whose `.env` path is resolved at import time.
- Cleanup lives in `pytest_unconfigure` because the existing `os._exit` there skips `atexit`, so a `tempfile`-registered cleanup would never run.

No test assertion was changed or weakened.

## Verification

Unit suite twice back to back in the same working tree, no manual cleanup:

```
===== RUN 1 =====
1521 passed, 36 warnings in 24.23s
===== RUN 2 =====
1521 passed, 36 warnings in 13.15s
```

`git status` is clean after both runs and no `.db` file is left in the tree. Collection count is unchanged versus pristine `main` (1521 both before and after), so nothing was skipped to get green. Integration suite also passes twice (41 passed, 26 skipped).

Lint gates: `ruff check src/` clean, `ruff format --check src/ tests/` clean. The 6 pre-existing `ruff check tests/` errors in three unrelated test files are untouched.


Sirajmx (approved):
> Verified the defect first, independently: on current `main` (without this fix), running
> `test_trust_tier_verification.py` twice back to back in the same working tree gets 24 passed,
> then `1 failed, 23 passed` on the rerun — the exact `AttributeError: 'NoneType' object has no
> attribute 'kwargs'` the PR describes, caused by the negotiation idempotency short-circuit
> replaying `ad_seller.db`'s cached response from run 1 and never invoking the mocked
> `counter_proposal`. Confirmed the fix on a clean merge onto current `main`: full unit suite twice
> back to back in one working tree, both green (1567/1567), `git status` clean and no `.db` file
> left in the tree either time. No test assertion was weakened to get there.
>
> Implementation checks out: the `DATABASE_URL`/`STORAGE_TYPE` override is module-level before the
> `ad_seller` import, which is required — `get_settings()` is `lru_cache`d and resolves its `.env`
> path at import time. Cleanup lives in `pytest_unconfigure` rather than `atexit` because the
> existing `os._exit` in that same function skips `atexit`, so a `tempfile`-registered cleanup would
> never run otherwise. Forcing `STORAGE_TYPE=sqlite` doesn't clobber anything: every test exercising
> a different backend (`test_storage.py`'s redis/mongodb cases) passes `storage_type` as an explicit
> function argument to `get_storage_backend(...)`, never reads it off the env/settings singleton.
>
> Cross-repo: zero risk by construction (only `tests/conftest.py` + `CHANGELOG.md`, no `src/` or
> wire-contract change). Also checked whether buyer-agent shares the same latent bug class, since its
> `Settings.database_url` defaults to the identical `sqlite:///./ad_buyer.db` working-tree pattern —
> it doesn't: ran its unit suite twice back to back on its own `main` (3512/3512 passed both times,
> no `.db` file created). Buyer-agent has ~15 separate per-entity store modules with no equivalent
> global lazy-singleton-through-settings pattern, so there's no shared on-disk state for a rerun to
> inherit. Approving.

PR #95, opened 2026-09-17
fix: stop leaking the seller's floor price and concession budget on negotiation reads

atc964:
## ⚠️ Security-relevant — please do not merge casually

This is an information-disclosure fix on a public reference implementation. It changes the response shape of a GET route and adds an auth requirement to it. Please review the wire-shape change and the auth decision deliberately rather than rubber-stamping.

## The defect

`GET /proposals/{proposal_id}/negotiation` had **no auth dependency at all** — no `Depends(...)`, no `response_model` — and returned whatever `negotiation_service.get_negotiation_status` produced. That projection was effectively a dump of the internal `NegotiationHistory`, including:

- `strategy` — the seller's negotiation posture for this buyer tier
- `base_price` — the seller's starting anchor
- `floor_price` — **the seller's absolute floor**
- `max_rounds` — the **remaining concession budget** (with `total_rounds` alongside it)

So any caller who knew or guessed a proposal id learned the seller's floor and how many concessions it had left. In a negotiation that is the one thing the seller must never reveal: a counterparty who knows the floor concedes nothing above it. Because this is a reference implementation, the leak was also a pattern others would copy.

These are exactly the four fields the shared `Negotiation` primitive excludes. From `RECONCILIATION.md` in `iab-agentic-primitives`: *"`floor_price`, `base_price`, `strategy`, and `NegotiationLimits` are deliberately excluded — they are the seller's internal guardrails and must never cross the wire."*

## What changed

**1. The guardrails are dropped at the service, not at the route.** `get_negotiation_status` now projects only buyer-visible facts (`negotiation_id`, `proposal_id`, `product_id`, `buyer_tier`, `status`, `total_rounds`, `rounds`, `started_at`, `completed_at`, `package_id`). Fixing it at the service rather than filtering at the route means no consumer of the projection can re-expose them. The only internal consumer — `apply_terminal_action` → `contract_mappers.terminal_round_response` — reads just `negotiation_id` and `rounds`, so nothing needed the removed fields.

**2. The wire shape is pinned.** New `NegotiationStatusResponse` is the route's `response_model`, so a later addition to the history dict cannot silently reach the wire again.

**3. The route requires a verified buyer.** It now resolves a verified buyer context exactly the way its siblings do (`POST /proposals/{proposal_id}/counter`, `POST /api/v1/negotiations/messages`) via `deps._get_optional_api_key_record` + `deps._verified_buyer_context`. A GET carries no body identity to self-assert, so an unverifiable caller cannot be established as a party to the negotiation and gets `401` instead of another buyer's negotiation state.

## Operator impact

Operators on an earlier build should assume floor prices and concession budgets were readable for every negotiation that existed on that build. Called out explicitly in the CHANGELOG for that reason.

**Breaking for any client reading the removed fields.** That is the point of the fix, but it is a wire change: clients reading `strategy` / `base_price` / `floor_price` / `max_rounds` off this route will stop seeing them, and anonymous readers will now get `401`. No buyer-side consumer in this repo reads them.

## Known gap, deliberately not fixed here

**The response is authentication-scoped, not buyer-scoped.** An authenticated buyer can still read *any* proposal's negotiation. `NegotiationHistory` records no buyer identity — only `buyer_tier` — so there is nothing to compare the requesting buyer against. The nearest candidate, `ProposalRequest.buyer_id`, is an optional self-asserted body field in a different namespace from `BuyerContext.get_pricing_key()`, and is absent entirely on quote-led negotiations. Real scoping needs a stored party identity on the history, which is a schema change beyond this defect. Recorded in the CHANGELOG rather than half-built.

## Tests

`tests/unit/test_negotiation_status_disclosure.py` (4 tests):

- the response carries none of `strategy` / `base_price` / `floor_price` / `max_rounds`, asserted against both the parsed body and the raw response text, plus the literal floor and base values so they cannot reappear under another key
- the **service projection** itself omits them, so a future consumer cannot re-expose them
- an anonymous caller gets `401` and no negotiation data in the body
- an authenticated caller still gets `200` (the gate is authentication, not a blanket denial)

**Negative control:** with `src/` reverted to `origin/main` (leak restored), 3 of the 4 fail — the guardrail assertion, the service-projection assertion, and the `401` assertion, each reporting the actual leaked values. The 4th (authenticated caller gets 200) correctly still passes, since the unguarded route answered everyone. The fix was then re-applied and the restored diff verified byte-identical to the pre-revert diff (matching sha256).

Full unit suite: **1525 passed**. `ruff check src/` and `ruff format --check src/ tests/` both clean.

## Sibling leaks found during the sweep — reported, not fixed

This class of leak does not appear once. Out of scope for this PR; flagging so the extent is known.

**Same field, other surfaces:**
- `GET /products` and `GET /products/{product_id}` are unauthenticated and return the shared `Product` with `ext.floor_cpm` populated from the product's floor (`contract_mappers.py:504`). Same information class, wider exposure, and the shared `Package` dropped `floor_price` for this exact reason.
- `AuthenticatedPackageView` (`models/media_kit.py:250`) exposes `floor_price` to any authenticated buyer.
- The MCP surface returns `floor_cpm` per product from `list_products` (`mcp_server.py:388`) and the configured default floor in setup status (`mcp_server.py:237`, `:314`).
- The chat adapter states the floor outright in its walk-away message: *"Our floor for this inventory is $X CPM"* (`chat/main.py:544`). Possibly an intentional product decision, but it is a floor disclosure.
- `GET /api/v1/rate-card` is unauthenticated while `PUT` on the same path is operator-gated, so the operator's stored base CPMs are world-readable. `RECONCILIATION.md` marks the rate card as never public.

**Assessed and cleared:**
- The `NEGOTIATION_STARTED` event payload does carry `strategy` and `base_price` (`negotiation_service.py:~260`), but it is **not externally visible**: `GET /events` and `GET /events/{event_id}` both require an operator key, and the event module has no webhook, HTTP, or external-broker sink. Operator-only.
- `deal_service.py:891` emits a `floor_price` key in the Xandr deal export, but the value is the deal's agreed price — an ad-server field name, not the seller's internal floor.
- Derived inference remains possible and is contract-sanctioned: `rounds[].cumulative_concession_pct` together with `seller_price` implies `base_price`, and `rounds_remaining` on the counter response implies `max_rounds`. Both fields are part of the shared `NegotiationRound` / `NegotiationRoundResponse`, so narrowing them would be a contract change, not a fix.

**Unauthenticated routes worth a deliberate public/private ruling** (many are legitimately public — health, agent card, catalog): `GET /api/v1/rate-card`, `GET /api/v1/deals/export`, `GET /api/v1/deals/{deal_id}`, `GET /api/v1/deals/{deal_id}/performance`, `GET /api/v1/deals/{deal_id}/lineage`, `GET /api/v1/quotes/{quote_id}`, `GET /sessions`, `GET /sessions/{session_id}`, `POST /sessions/{session_id}/messages`, `POST /sessions/{session_id}/close`, `POST /deals`, `POST /api/v1/deals/curated`, `POST /api/v1/deals/{deal_id}/deprecate`, `GET /registry/agents`, `GET /registry/agents/{agent_id}`. The deal/quote/session reads are the same read-anyone's-record-by-id shape as the defect fixed here.


Sirajmx (approved):
> Confirmed the defect live on current `main` before touching the fix: the route has no
> `Depends(...)` and no `response_model` at all, so it returns the raw service projection —
> `strategy`, `base_price`, `floor_price`, `max_rounds` included.
>
> Ran a real negative control myself, not just the PR's claimed numbers: mounted the new test file
> into a container running current `main` (without the fix) and ran it directly. 5 of 6 tests
> genuinely fail — guardrail fields present in both the route response and the service projection,
> anonymous caller gets `200` not `401`, and the engine-driven round dump concretely shows
> `concession_pct`/`cumulative_concession_pct`/`rationale` on the wire. That last one matters most:
> checked the reconstruction claim mathematically (`seller_price / (1 - cumulative_concession_pct)`
> does exactly reconstruct `base_price`), confirmed via the test's own sanity assertion, so a fix
> that only dropped the top-level fields while serializing rounds raw would have been a no-op.
>
> Fix verified on a clean merge onto current `main`: all 6 new tests pass, full suite 1573 passed,
> ruff clean. The typed `NegotiationRoundView` whitelist genuinely closes the round-level leak —
> verified only the 5 allowed fields ever appear on the wire.
>
> Cross-repo: zero impact, confirmed by grep — buyer-agent never calls this route and doesn't read
> `strategy`/`floor_price`/`max_rounds`/`cumulative_concession_pct` anywhere; the `base_price`/
> `floor_price` hits that do exist there are unrelated (product catalog pricing, a different
> connector's deal object).
>
> The PR is upfront about its own remaining gap (authentication-scoped but not buyer-scoped — any
> authenticated buyer can still read any proposal's negotiation, since `NegotiationHistory` records
> no buyer identity to scope against) rather than half-building buyer scoping into this diff. Its
> sweep of sibling leaks (unauthenticated `/products`, the rate card, MCP `list_products`, the chat
> floor disclosure) checks out and is exactly what #97 and #98 in this same batch then go fix.
> Breaking wire change + new 401 justifies the security warning on this one — reviewed deliberately
> rather than rubber-stamped. Approving.

PR #97, opened 2026-09-17
fix: stop disclosing the seller's floor price in chat and on every counter-offer

atc964:
## Security-relevant — please do not merge casually

This is an information-disclosure fix on a public reference implementation, and it **removes a field from a response body on the normal negotiation path**. Please review the wire change deliberately rather than rubber-stamping.

**This is the second of at least two PRs in this class.** #95 fixed the same defect class on `GET /proposals/{proposal_id}/negotiation`. That PR's own sweep found more surfaces, and this PR fixes the two that are worst by reachability. Others remain — see "Still open" below. #95 and this PR are independent and can merge in either order.

## The defect

Two surfaces disclosed the seller's floor price, and neither is an obscure route.

**1. The chat interface said it out loud.** `interfaces/chat/main.py:544`, on a negotiation REJECT, emitted the literal sentence:

> Our floor for this inventory is **$33.11 CPM**. Would you like to explore other packages that might fit your budget?

formatted from `history.floor_price`. The seller told its counterparty its own floor, in prose, mid-negotiation.

**2. Every counter-offer carried it.** `flows/proposal_handling_flow.py:821` put `"floor_price": product.floor_cpm` into `state.counter_terms`. `counter_terms` is returned to the buyer on **every** counter-offer (it is where the buyer reads `proposed_price` and `negotiation_id`), so this crossed the wire on the normal negotiation path — not on a read someone had to go looking for. Observed payload before the fix:

```
{'proposed_price': 33.11, 'floor_price': 33.11, 'max_impressions': 1000000,
 'reason': '...', 'negotiation_id': 'neg-2602e115', 'round_number': 1, 'action': 'counter'}
```

A counterparty who knows the floor concedes nothing above it, so either leak makes the seller's negotiating position structurally worthless. `floor_price`, `base_price`, `strategy` and `NegotiationLimits` (`max_rounds`) are the four fields the shared `Negotiation` primitive excludes deliberately — from `RECONCILIATION.md` in `iab-agentic-primitives`: *"`floor_price`, `base_price`, `strategy`, and `NegotiationLimits` are deliberately excluded — they are the seller's internal guardrails and must never cross the wire."*

Because this is donated reference code, the pattern propagates to every implementer who copies it.

## What changed

**1. The chat REJECT is now a walk-away with no number in it.** House voice kept, consistent with the neighbouring FINAL_OFFER branch, and it still offers the buyer a next step.

It also **no longer interpolates the engine's REJECT rationale**. That rationale has a max-rounds variant reading *"Maximum 5 rounds reached"*, which publishes the remaining concession budget in prose. Dropping the interpolation is local to the branch being rewritten and removes a second guardrail disclosure from the same message.

**2. `floor_price` is removed from `counter_terms`.** Nothing else in the dict changed.

**3. Both sites carry a comment saying why**, naming the four fields, so the next person to extend either payload sees the constraint at the point of edit.

## Cross-repo check before removing the field — done first, not after

`counter_terms` is read by the buyer agent, so removing a key from it could have broken the buyer. Checked against `IABTechLab/buyer-agent` at **`origin/main`** before touching it:

- `ad_buyer/orchestration/multi_seller.py:1422-1434` reads exactly two keys: `counter_terms["proposed_price"]` and `counter_terms["negotiation_id"]`. Both untouched.
- **No buyer code reads `counter_terms["floor_price"]`.** No hits anywhere in `src/`.
- The buyer's only `floor_price` consumer is `ad_buyer/media_kit/client.py:110`, in `_parse_package_detail`, which parses `GET /media-kit/packages/{id}`. That is a **different seller surface** (`AuthenticatedPackageView.floor_price`) and is deliberately **not** touched here — removing it would break the buyer's media-kit client, and it needs the public-versus-private ruling described below.

So this is safe for the buyer as it stands. A test in the new file pins the two keys the buyer does read, so a later tidy-up of `counter_terms` cannot quietly remove them.

## Tests

`tests/unit/test_counter_and_chat_floor_disclosure.py` — 11 tests in three sections.

Sections 1 and 2 pin the two sites: the walk-away contains neither the floor value, nor the base price, nor the word "floor", nor the round budget, and is still a usable message; `counter_terms` has no `floor_price` key and still carries what the buyer reads.

Section 3 is **the part that matters for the next leak**: a by-field-name sweep. It recurses to any depth and asserts that neither path carries `floor_price`, `base_price`, `strategy` or `max_rounds` **as a field name** — over `counter_terms`, over the whole buyer-facing proposal response, and over the chat negotiation envelope. A per-site test catches today's two leaks; this catches the next guardrail someone adds to either payload without anyone having to remember to write a new test. One test deliberately asserts that the internal `NegotiationHistory` *does* contain all four, so the sweep is proven non-vacuous rather than passing because it is looking at nothing.

This **complements** #95 rather than duplicating it: that file pins one route's response shape and its auth gate; this one pins the chat text and the counter payload.

**Negative control.** With both leaks restored from `origin/main`, **6 of the 11 fail**, each reporting the actual leaked value — the floor `33.11` in the chat text, `Maximum 5` for the round budget, and `['floor_price']` in the `counter_terms` payload. The 5 that still pass correctly should: the walk-away-usability test, the buyer-keys test, the non-vacuousness test, the chat envelope test (that envelope never leaked), and the name sweep over the walk-away text — prose says "our floor", not `floor_price`, which is exactly why sections 1 and 2 assert on the value and the wording rather than on the field name. The fix was then re-applied and the restored diff verified byte-identical to the pre-revert diff (matching sha256).

Full unit suite: **1532 passed**. `ruff check src/` and `ruff format --check src/ tests/` both clean.

## Known remaining gap on this class — recorded, not half-built

**The engine's `rationale` strings still state the floor and the strategy in prose.** From `engines/negotiation_engine.py`:

- `"Buyer offer $25.00 is below our floor. Countering at the floor price $33.11 CPM — the minimum viable price for this inventory (round 1/5)."` — states the floor **value**, and the round budget
- `"Counter at $30.00 CPM (collaborative strategy, round 1/5)."` — names the **strategy** and the round budget
- `"Final offer at our floor price $X CPM. We cannot go lower."` — labels the price as the floor
- `"...our maximum concession of 15%."` — the concession cap

These reach the buyer through `counter_terms["reason"]`, so **the floor is still derivable in prose on the counter path** even with this fix applied. A leak moved from a named field into a prose string is not fixed, and this PR does not claim to have fixed it.

It is not fixed here because the strings are built in the negotiation engine, a different surface from the two fixed in this PR, and rewriting how the engine phrases every rationale is a behavioural change to negotiation output that deserves its own review — not a side effect of a disclosure patch. Sanitising the prose at the flow boundary instead would be lossy guesswork on free text. Called out in the CHANGELOG and in the test module's docstring so the next reader is not misled into thinking the floor is fully off the wire.

## Still open on this class, deliberately out of scope

Each of these needs a deliberate public-versus-private ruling rather than a reflexive patch, and one of them (the package view) has a live buyer consumer that a naive removal would break:

- `GET /products` and `GET /products/{id}` are unauthenticated and return `Product.ext.floor_cpm` (`contract_mappers.py:504`)
- `AuthenticatedPackageView.floor_price` (`models/media_kit.py:250`) exposes a package floor to any authenticated buyer — **and the buyer agent's media-kit client reads it**
- the MCP surface returns `floor_cpm` per product (`mcp_server.py:388`) and a default floor in setup status (`:237`, `:314`)
- `GET /api/v1/rate-card` is unauthenticated while `PUT` on the same path is operator-gated, though `RECONCILIATION.md` marks rate cards as never public

## Operator impact

Operators running an earlier build should assume their product floors were readable by **any buyer who received a counter-offer**, and by **any buyer who was rejected in chat**. Both are ordinary negotiation outcomes, so the realistic exposure is "every buyer who negotiated", not "someone who probed an unusual endpoint". Called out in the CHANGELOG for that reason.

No version numbers changed.


Sirajmx (approved):
> Reproduced both leaks live, independently, on current `main`: the chat REJECT branch genuinely
> outputs "Our floor for this inventory is **$33.11 CPM**... Maximum 5 rounds reached..." — floor
> value and round budget both present, built from a hand-constructed history/round, not the PR's
> own fixture. Verified the fix live on a clean merge onto current `main`: the same inputs through
> the fixed formatter now return "...we've gone as far as we can on price here..." — no floor
> value, no word "floor" anywhere. New test file: 25 passed. Full suite: 1592 passed. Ruff clean.
>
> One thing checked hard before approving, since it looked like a possible miss: the shared
> `NegotiationRound` primitive (used by the canonical `POST /api/v1/negotiations/messages` route)
> genuinely declares `concession_pct`/`cumulative_concession_pct` as real fields — confirmed via
> `NegotiationRound.model_fields` and that `extra: ignore` doesn't drop them, so they do reach that
> wire, the same base_price-reconstruction math #95 fixed elsewhere. Not a miss in #97, though —
> #95's own body already names this exact gap and calls it "genuinely contract-sanctioned...
> narrowing those is a contract discussion, not a fix here" (shared-library spec issue, correctly
> out of scope for either PR). #97's actual job on this route, the `rationale` string content, is
> correctly fixed — `counter_proposal()` now populates it from `buyer_rationale`, so the leaked
> field name is a known/disclosed gap, but the prose itself is safe.
>
> Cross-repo, re-verified independently rather than trusting the PR's own claim: `multi_seller.py`
> reads exactly `counter_terms.get("proposed_price")` and `.get("negotiation_id")`, nothing else.
> The other `floor_price` reads in buyer-agent are unrelated — a different connector's raw deal
> object, and `media_kit/client.py`'s genuinely separate `AuthenticatedPackageView` surface, which
> is exactly why the PR correctly leaves that one untouched. Approving.

PR #98, opened 2026-09-17
fix: stop the unauthenticated change-request listing from returning every order's audit trail

atc964:
## ⚠️ Security-relevant — please do not merge casually

Information disclosure on a public reference implementation. This adds an auth requirement to three routes, scopes their results to the caller, changes the response shape of all five, and removes a field from a request body. Please review the wire change and the auth decisions deliberately rather than rubber-stamping.

Fixes `seller-928` (P0).

## The defect

`GET /api/v1/change-requests` took **optional** auth (`_get_optional_api_key_record`, which returns `None` for anonymous callers), required **no filter** (`order_id` and `status` are both `Optional`), and declared **no `response_model`**. With no query params the service built an empty filter and asked storage for everything.

Each record carries `rollback_snapshot`, written at `order_service.py:365` as `order.copy()` — the full order: `OrderStateMachine.to_dict()` (`order_id`, `status`, and the `audit_log` with every transition's **actor and reason**) plus `deal_id`, `quote_id`, `created_at` and `metadata`.

So one anonymous unfiltered GET returned every tenant's order audit trail. Here is the actual body, from the negative control below:

```
{"change_requests":[{... "rollback_snapshot":{"order_id":"ORD-VICTIM","status":"booked",
"deal_id":"DEMO-VICTIM","quote_id":"QT-VICTIM-SECRET","metadata":{"campaign":"victim-spring-2026"},
"audit_log":{"order_id":"ORD-VICTIM","transitions":[{"from_status":"draft","to_status":"booked",
"actor":"human:victim-trader@agency.example",
"reason":"renegotiated after Q3 shortfall call with CFO"}]}}}],"count":1}
```

`POST /api/v1/change-requests` was anonymous for the same reason and took `requested_by` straight off the wire (`schemas.py:364`, `requested_by: str = "system"`), writing it into the persisted record verbatim. Nothing downstream *decides* on `requested_by` — approval routing is severity-only and review/apply are operator-gated — so this is not privilege escalation. It is worse in the place that matters: `requested_by` is exactly what a human reviewer looks at when deciding whether to approve, so the practical attack is social-engineering the approval (`"requested_by": "human:cfo@publisher.example"`), and the audit trail that would catch it was the thing being disclosed.

Note the contrast on the same router: `review` and `apply` already used `_require_operator_api_key_record`. Only create and the two reads were open.

## What changed

**1. Authentication.** Create, list and read-by-id now take `require_api_key_record` (anonymous → 401) — the same dependency the approvals router uses. Review/apply keep their stricter operator gate. A **buyer** key is sufficient for the buyer-facing three; this is not a promotion to operator-only.

**2. Ownership scoping — because authentication alone is not enough.** An authenticated buyer would still read every other buyer's trail. Each record is stamped at creation with an actor derived from the credential, and a buyer caller sees only records carrying its own actor. An operator caller is unscoped: reviewing the queue is the seller's job and needs all of it.

The actor comes from a new `auth.dependencies.actor_from_api_key`. It deliberately differs from the existing `principal_from_api_key`: that one anchors to the immutable `key_id` because an approval audit entry must name the exact credential that decided, whereas this is an **ownership key** and must survive key rotation, or a buyer would lose sight of records it filed under an earlier key. So it anchors to the strongest stable buyer identifier (`agent:<advertiser|agency|seat>`), falls back to `agent:key:<key_id>` only for a credential asserting no buyer identity at all, and gives operators their own `operator:` namespace so a buyer identifier can never collide with a seller-side actor.

**Another actor's record reads `404`, not `403`.** A distinguishable "exists but not yours" would turn the by-id route into an existence oracle over every tenant's change request ids.

**3. The wire shape is pinned.** All five handlers declare a `response_model` (`ChangeRequestResponse`, `ChangeRequestListResponse`, `ChangeRequestApplyResponse`). `rollback_snapshot` is absent from those models by construction, so it cannot reach the wire even if a future consumer serialises the record wholesale.

**4. The actor is stamped, not asserted.** `requested_by` is removed from `CreateChangeRequestModel` and derived from the credential in the router. A body that still sends it is accepted and the value ignored (pydantic drops unknown fields), so this is not a hard break for existing clients — but the value they send no longer does anything.

**5. Idempotency keys are namespaced per actor.** The key was `idempotency:change-request:{order_id}:{idempotency_key}`. Without the actor in it, one buyer replaying another's `(order_id, idempotency_key)` pair would be handed back the other buyer's change request through the replay path, straight past the new scoping. Pinned by a test.

**6. The list scope is applied in the service, in Python — not in the storage filter dict.** `StorageBackend.list_change_requests` understands only `order_id` and `status` and silently ignores any other key (`storage/base.py:260-274`), so a `requested_by` pushed in that way would **fail open** and return every tenant's records. There is a test asserting the scope is not passed that way.

## Deliberately NOT done here

**`rollback_snapshot` is still written.** It is genuinely write-only — a repo-wide grep over `src/` finds exactly two references, the declaration and the one write, and `apply_change_request` never reads it. Deleting it is the right call and is the cheapest part of the fix, but it is a persisted-storage-shape change (existing records carry it; both storage backends should be confirmed to deserialize them) and that is beyond the disclosure. This PR closes the wire completely — `rollback_snapshot` appears in no response on this router — and the deletion is tracked as **`seller-d4x`** (P2).

## Tests

`tests/unit/test_change_request_disclosure.py` (16 tests), in four groups: the auth gate, the ownership scoping, the declared wire shape, and the server-stamped actor. Notable ones:

- the anonymous listing is `401` **and** the response text contains none of `rollback_snapshot`, the order id, the secret quote id or the audit-log reason — asserted against raw text, so the values cannot reappear under another key
- an authenticated *stranger* gets an empty list and a `404` by id, while the owner still gets `200` and its own record (the gate is scoping, not a blanket denial)
- an operator still sees the whole queue
- replaying another buyer's idempotency key does not return that buyer's record
- **all five** handlers are driven end to end and none emits `rollback_snapshot`, the audit log, or the quote id — plus a check that the snapshot *is* still in storage, so the test pins the wire rather than the storage shape
- every route object on the router has a non-`None` `response_model`
- a body `requested_by` of `human:cfo@publisher.example` does not appear in the response or in the persisted record
- the actor is stable across key rotation, operator actors cannot collide with a buyer identifier, and two identity-less buyer keys do not collapse onto one shared actor
- the service-level scope is not pushed into the storage filter dict

**Negative control:** with `src/` reverted to `origin/main` (and a local shim standing in for the new `actor_from_api_key`, which is new code and cannot exist on the old tree), **11 of the 16 fail**, reporting the actual leaked values shown above. The 5 that pass are the 3 pure helper unit tests plus "owner still sees its own records" and "operator sees the whole queue" — both of which correctly still pass, since the unguarded route answered everyone. `src/` was then restored and the reapplied diff verified byte-identical to the pre-revert diff (matching sha256).

Full unit suite: **1537 passed**. `ruff check src/` and `ruff format --check src/ tests/` both clean.

> One note on the local run: `tests/unit/test_trust_tier_verification.py::TestNegotiationMessagesCeiling::test_self_asserted_advertiser_identity_is_floored` fails when the suite runs against a *reused* sqlite file, because a stale idempotency row short-circuits the endpoint. It fails identically on a pristine `origin/main` checkout, so it is pre-existing and unrelated — it is the worktree `.env` inheritance already tracked as `seller-1jw`. With an isolated `DATABASE_URL`, 1537/1537 pass.

---

# The 16-route ruling

PR #95 listed 15 unauthenticated routes and asked for a deliberate public/private ruling. One question decides all of them:

> **A route is legitimately public only if what it returns is identical for every caller and is content the seller intends to publish** — health, the agent card, the catalog. Anything keyed by a per-counterparty record id (a deal, a quote, a session) fails that test: the id is the only thing standing between an outsider and another buyer's commercial terms, and **an id is not a credential**. Anything that changes state is not a read at all.

By that rule, **zero** of the sixteen are legitimately public.

| # | Route | Ruling | Reasoning |
|---|---|---|---|
| 1 | `GET /api/v1/change-requests` | **(b)** — **fixed in this PR** | Unauthenticated, unfiltered, no response model: every tenant's order audit trail in one request. |
| 2 | `GET /api/v1/deals/export` | **(b)** operator | Walks `deal_index` and returns **every** deal with per-deal prices, no filter required. The seller's whole book, anonymously. Sibling `/deals/push` and `/deals/distribute` are already operator-gated. |
| 3 | `GET /sessions` | **(b)** + scope | Unauthenticated and unfiltered: every buyer session with `buyer_pricing_key`, `negotiation_stage`, `message_count`. A cross-tenant roster of who is negotiating and how far along. |
| 4 | `GET /sessions/{session_id}` | **(b)** + scope | Full conversation transcript, negotiation object and `buyer_pricing_key` for any session id. Sessions already persist the buyer key, so scoping needs no schema change. |
| 5 | `POST /sessions/{session_id}/messages` | **(b)** + scope | Not a read. Injects a message into another buyer's live negotiation and returns the agent's reply — and with #3 open the session ids are not even secret. |
| 6 | `POST /sessions/{session_id}/close` | **(b)** + scope | Not a read. Terminates another buyer's session. |
| 7 | `GET /api/v1/quotes/{quote_id}` | **(b)** + scope | Read-anyone's-quote-by-id, including price. `QuoteHistoryStore` already stores `buyer_id` — the booking path at `deals.py:~185` compares it against `context.get_pricing_key()`, so the seller **can** tell whose quote it is; the GET simply doesn't ask. |
| 8 | `GET /api/v1/deals/{deal_id}` | **(b)** + scope | Read-anyone's-deal-by-id: terms and price. Same shape as the defect fixed here. |
| 9 | `GET /api/v1/deals/{deal_id}/performance` | **(b)** + scope | Another party's delivery stats. Mock data today, real ad-server data by design later — gate it before it carries real numbers, not after. |
| 10 | `GET /api/v1/deals/{deal_id}/lineage` | **(b)** + scope | Walks parents and replacements, so it maps a counterparty's deal history from one id. |
| 11 | `POST /deals` (legacy `generate_deal`) | **(b)** | Not a read. Anonymous deal generation from a proposal id, while its modern sibling `POST /api/v1/deals` is credential-gated and verifies the buyer against the quote. A duplicate surface with weaker auth is the one an attacker uses. Gate to parity or delete. |
| 12 | `POST /api/v1/deals/curated` | **(b)** | Not a read. Anonymous creation of a priced deal carrying a self-named curator, its fee, and a `schain` node — while registering a curator (`POST /api/v1/curators`) is already operator-gated. |
| 13 | `POST /api/v1/deals/{deal_id}/deprecate` | **(b)** | Not a read. Anyone can mark **any** deal deprecated and point its "replacement" wherever they like; the buyer's Deal Library reads lineage to decide what to stop targeting, so this is a remote off-switch for someone else's live deal. |
| 14 | `GET /api/v1/rate-card` | **(c)** human call | Public `GET` while `PUT` on the same path is operator-gated — the same document is confidential to write and public to read. Since #69 a stored entry **overrides catalog base CPM** as the live price anchor, and `RECONCILIATION.md` marks the rate card as never public. Needs a level ruling because the handler serves generic `source:"defaults"` values (which do **not** drive pricing) and the operator's `source:"stored"` card from one path; only the latter is a disclosure. |
| 15 | `GET /registry/agents` | **(c)** human call | The only genuinely arguable pair. A registry of known agents and their trust status is plausibly public infrastructure — but `trust_status` is the **seller's private assessment** of a counterparty (including `blocked`), and publishing "we have blocked this agent" is a business disclosure, not a directory entry. Either redact `trust_status` on the public read or gate the route; that is a product call. |
| 16 | `GET /registry/agents/{agent_id}` | **(c)** human call | Same object, same question. Decide once with #15. |

**Nothing in the list is (a).** The genuinely public surfaces — `GET /health`, `GET /`, `GET /.well-known/agent.json`, the product catalog — were not on it, which is itself the signal: the list is 15 routes that each return or mutate something specific to one counterparty.

### Why none of the other 15 is fixed in this PR

The brief was to fix the clear `(b)` cases here if they were small and mechanical. On inspection none is. Every one needs either a **level decision** the code cannot make (buyer key vs operator key: is deprecating a deal the buyer's act or the seller's?) or an **ownership key** to scope against — and for deals that key may not be stored at all, which is the same gap PR #95 recorded for `NegotiationHistory` and declined to half-build. Adding five more surfaces to a P0 disclosure fix would also bury the reviewable part of this diff. They are filed instead:

- **`seller-c5r`** (P0) — rows 2, 3, 4, 7, 8, 9, 10: the unauthenticated reads that hand back another party's record, or every party's records.
- **`seller-h8d`** (P0) — rows 5, 6, 11, 12, 13: the five state-changing routes that take no credential at all.
- **`seller-3p0`** (P1) — row 14: the rate card, with the three candidate rulings spelled out.
- **`seller-d4x`** (P2) — deleting `rollback_snapshot` itself.

Rows 15 and 16 are a product call rather than an engineering defect and are left in this table for a human to rule on rather than filed as a bug.


Sirajmx (approved):
> The most serious of this batch (P0, `seller-928`). Reproduced the defect live and independently —
> not via the PR's own fixture, a hand-built order + change request through the real service layer
> — and the pre-fix listing genuinely returns the full `rollback_snapshot`, near-exact match to the
> PR's own example: the victim's `quote_id`, campaign metadata, and audit-log actor/reason all
> present, unauthenticated.
>
> Fix verified on a clean merge onto current `main` (one trivial CHANGELOG conflict from #94
> landing since I last checked, resolved by concatenation): new disclosure test file 16/16 passed,
> existing `test_change_request.py` + `test_operator_auth.py` 103 passed (the scoping changes don't
> break owner/operator behavior), full suite 1583 passed, ruff clean.
>
> Checked the design choices, not just read them: 404-not-403 for another actor's record is right
> (a 403 would let status codes enumerate valid ids); scoping applied in Python rather than pushed
> into the storage filter dict is the safe choice — confirmed `StorageBackend.list_change_requests`
> silently ignores unknown filter keys, so the naive approach would fail open; the idempotency key
> namespaced per actor closes a real replay hole (without it, one buyer replaying another's
> `(order_id, idempotency_key)` pair would get handed the other buyer's record); and
> `actor_from_api_key` correctly anchors to buyer identity rather than `key_id`, since an ownership
> key must survive key rotation while an audit identity should not.
>
> Cross-repo: zero impact, confirmed by grep — buyer-agent has no change-request client code at
> all, no hits for `change-request`, `requested_by`, or `rollback_snapshot` anywhere in its `src/`.
>
> `rollback_snapshot` itself is still written to storage, deliberately left for a follow-up since
> deleting it is a persisted-storage-shape change, not a wire-disclosure fix — confirmed it's
> genuinely write-only (exactly two references repo-wide) so that follow-up is low-risk whenever it
> lands. Approving.

PR #99, opened 2026-09-18
fix: MediaKitService() crashes list_packages and silently breaks get_setup_status (AI-12)

Sirajmx:
## The defect

The MCP `list_packages` tool constructs `MediaKitService()` with no arguments, but the class requires `storage`/`pricing_engine` positionally. Every call raises unconditionally:

```
> service = MediaKitService()
TypeError: MediaKitService.__init__() missing 2 required positional arguments: 'storage' and 'pricing_engine'
```

`get_setup_status` has the identical bug at its own media-kit check, but wrapped in a bare `try/except: pass`, so the crash never surfaces — it just silently reports `media_kit.configured: false` no matter what's actually in storage. Confirmed live: created a real `active` package via REST (`POST /packages`), then called `get_setup_status` over MCP — it still reported `{"configured": false, "package_count": 0}`, even though the package genuinely existed.

The equivalent REST routes (`GET /packages`, `GET /media-kit/packages`) construct the service correctly and are unaffected — this is specific to these two MCP call sites.

Already filed as `AI-12` (Highest priority — blocks the publisher happy path), reported independently of this fix.

There's a pre-existing integration test that targets exactly this — `tests/integration/test_mcp_streamable.py::test_list_packages` — but it self-skips silently unless a live server is already running and `SELLER_MCP_HTTP_URL` is set, so a bare `pytest tests/integration/` run reports it as "skipped," not failed. Ran it against a real running server on pristine `main`: it fails with the exact `TypeError` above, confirming this test would have caught the regression if it were ever actually exercised against a live server in CI.

## The fix

Added one `_media_kit_service()` helper that builds `MediaKitService` with `storage`/`pricing_engine`, mirroring the existing pattern `_registry_service()`/`_api_key_service()` already use in this same file (construct the same way `interfaces.api.deps` does, without importing the app module). Both call sites — `list_packages` and `get_setup_status` — now go through this one helper instead of each re-instantiating the service independently, so there's a single place this can break again instead of two.

## Behavior change

`get_setup_status`'s `media_kit.configured`/`package_count` fields now reflect reality instead of always reporting empty. Operators who relied on the setup wizard's media-kit check should re-run it — a "configured" media kit may now correctly show as configured for the first time.

## Verification

New tests in `tests/unit/test_mcp_server.py`: `TestMediaKitServiceHelper` (constructs the real service against a mocked storage, asserts the actual type/args) and `TestListPackages` (drives the tool end to end). Both deliberately avoid the existing test file's own pattern of mocking out the whole `media_kit_service` module — that pattern is exactly why this bug was invisible to the existing suite.

Negative control: reverted both call sites back to `MediaKitService()`, both new tests failed as expected (one `NameError`, one crash surfaced through the tool), confirming they're not vacuous.

Full unit suite: **1523 passed** (1521 on `main` + 2 new). Full integration suite (`tests/integration/`, matching CI's exact invocation): **53 passed, 13 skipped, 1 failed** — the one failure (`test_get_inbound_queue`, an auth-drift issue unrelated to media kit) is pre-existing on `main` too; confirmed identically on both branches, not introduced here.


atc964 (approved):
Constructor fix is correct and complete. Both zero-arg sites covered, no others in the repo, and the new tests genuinely fail pre-fix.

#106 fixes a different term of the same expression. `setup_complete` is `all([identity_configured, ad_server_configured, len(packages) > 0])`. This PR makes the third term reachable, since `MediaKitService()` raised unconditionally and the exception was swallowed, so `packages` was always empty. #106 makes the second term true for CSV and S3. In CSV mode both must land before `setup_complete` can become true. They merge clean and the order does not matter.

**Follow-up, not blocking:** the title says "silently breaks `get_setup_status`", and the silence is still there. The `except Exception: pass` is untouched, with no logging, in a file that has a module-level logger. A malformed package row or a storage outage still renders as "Setup is incomplete. I can walk you through the setup wizard step by step." That predates this PR, so it belongs in its own change, but the title currently claims more than the diff delivers.

PR #100, opened 2026-09-18
fix: inventory sync now persists what it fetches, and stops hardcoding a $10 CPM floor fallback (AI-8, AI-9, AI-10)

Sirajmx:
## The defect

`inventory_sync_scheduler._run_sync()` — the function behind both the scheduled background sync and the MCP `sync_inventory` tool — called the ad server client directly, fetched real inventory items, counted them, and discarded them:

```python
client = get_ad_server_client()
async with client:
    items = await client.list_inventory(filter_str=filter_str)
return {"status": "success", "items_synced": len(items), "synced_at": _last_sync}
```

`items` is never used for anything but `len(items)`. No product is created, no package is written to storage. A "successful" sync leaves the seller's inventory exactly as empty as before it ran (AI-9), and creates zero packages (AI-10) — confirmed live: called `sync_inventory` against a fresh instance, got back `{"status": "success", "items_synced": N}`, and `GET /packages` still showed 0 afterward.

Separately, wherever a synced item carries no `floor_price_cpm` from the ad server (the common case for real GAM/FreeWheel items), the fallback was a hardcoded `10.0` in two places — `catalog_service.product_from_inventory_item` (feeds product `base_cpm`) and `ProductSetupFlow._estimate_base_cpm` (feeds synced-package `base_price`/`floor_price`) — silently ignoring the operator's configured `default_price_floor_cpm` (AI-8). Confirmed live: a CSV item with no floor data, `default_price_floor_cpm=22.5`, priced at a hardcoded **$10.00** instead.

All three already filed (AI-8, AI-9, AI-10 — High priority, "major incoherence"), reported independently of this fix.

## The fix

**AI-9/AI-10:** `_run_sync()` now calls `ProductSetupFlow().kickoff_async()` instead of re-deriving a thinner version of the same thing. The flow already does this correctly — idempotent upsert of SYNCED-layer packages plus pruning of stale ones — and is the exact path `POST /packages/sync` already uses in production, so this reuses a proven implementation instead of duplicating persistence logic in a second place. `include_archived` is kept as a parameter for backward compatibility, but the flow doesn't currently support filtering by archived status — neither does `POST /packages/sync`, which has the identical pre-existing limitation. Not fixed here; a real but separate, smaller gap.

**AI-8:** both fallback sites now read `settings.default_price_floor_cpm` instead of a literal `10.0`. The per-type CPM estimates in `_estimate_base_cpm` (display=$12, video=$25, etc.) are untouched — only the catch-all fallback for an unrecognized type changes.

## Verification

New test file `tests/unit/test_inventory_sync_persistence.py` (6 tests): `_run_sync()` actually persists packages and returns counts that match what's in storage, a repeated sync stays idempotent, and both floor-fallback sites respect the configured value while leaving an item's own explicit floor and the per-type estimates untouched.

Negative controls: reverted `_run_sync()` to the old discard-everything version — the two persistence-specific tests failed as expected (the idempotency test still passed, trivially, since 0 == 0). Reverted both floor fallbacks to the hardcoded `10.0` — both floor tests failed as expected.

While verifying, found and fixed a real pre-existing gap in `tests/unit/test_issue34_catalog_fixes.py`'s own `_flow_settings()` test helper: it didn't define `default_price_floor_cpm`, so once `_estimate_base_cpm` started reading it, `ProductSetupFlow.sync_from_ad_server()`'s broad `except Exception` silently caught the resulting `AttributeError` and fell back to mock packages — masking the real cause as "ad server sync failed." Added the missing field to the fixture rather than working around the masking.

Full unit suite: **1527 passed** (1521 baseline + 6 new), 0 failures. Full integration suite (`tests/integration/`): 39 passed, 28 skipped (matches baseline shape without a live server), 0 failures. `ruff check`/`ruff format --check`: clean.

Live, end-to-end, on two separately running instances (pristine `main` vs. this branch): called `sync_inventory` over real MCP — `main` returns `{"status": "success", "items_synced": N}` with `GET /packages` still at 0 afterward; this branch returns the same plus `"packages_synced": 4`, and `GET /packages` shows all 4 real, persisted packages. For AI-8: a CSV-mode instance with no `floor_price_cpm` in the data and `default_price_floor_cpm=22.5` configured on both — `main`'s `GET /products` prices the item at $10.00 regardless; this branch prices it at $22.50.


atc964 (changes requested):
Two of these move real numbers, and one turns an outage into data loss. Holding on those rather than the defaults question.

**1. This halves the floor for all GAM and FreeWheel inventory.** Those adapters never attach `raw`, so the fallback always fires: `base_cpm 10.00 / floor_cpm 8.50` becomes `5.00 / 4.25`. CSV is unaffected because all 35 sample rows carry `floor_price_cpm`, which is why the tests at 22.5 and 17.0 pass while never showing the direction it moves by default.

**2. The worst consumer is not price display, it is `check_avails`.** `requested = int(budget / cpm * 1000)`. Halving the CPM doubles the impressions we claim are available for a fixed budget, with `total_cost` unchanged. That is a delivery commitment, not a label.

**3. An ad-server outage becomes an unattended catalog wipe reported as success.** `await self._finish_sync()` sits outside the `try/except`, so the prune runs on the failure path, and `keep` holds only the four mock ids. Every real synced package is deleted. Nothing logs above INFO, and `get_sync_status` advances its timestamp and counter, so the endpoint shows positive evidence of a healthy sync after the catalog was replaced with dev fixtures. That mechanism predates this PR. What this PR changes is that it promotes it from an explicit operator action to a recurring hourly background job.

**4. `include_archived` is silently dropped for GAM.** `filter_str = None if include_archived else "status:ACTIVE"` is gone and `list_inventory()` is now called bare, so archived ad units sync into the live media kit. The flip side is worth stating too: this incidentally fixes CSV, where that filter was being applied as a name substring and matched zero rows.

**Required before merge:**

1. Gate the prune on the ad-server path having actually succeeded, and keep `_run_sync` returning `status: "error"` when the sync failed. As written, a transient GAM outage destroys the real synced layer.
2. Raise `default_price_floor_cpm` to `10.0` in this PR so the change is a true no-op. The value is read in three places, its definition and two status reports, so nothing else shifts.
3. Restore the `status:ACTIVE` filter for the GAM and FreeWheel path, or state explicitly in the PR body that dropping it is intended.

**Follow-up, not blocking:**

- "Persists what it fetches" covers packages but not products. There is no `set_product` call in the flow, so `items_synced: 47` can be reported with zero products reaching any buyer-queryable surface.
- `.get(inv_type, self._settings.default_price_floor_cpm)` evaluates the default eagerly on every call, including recognised types. That is why the settings fixture had to change, and any future fake-settings double will `AttributeError` on all types rather than just unrecognised ones.
- `default_price_floor_cpm` is consumed as a base, and the product floor is then derived at 85% of it while the package floor is 70%. So a publisher who raises their floor still sees synced products priced below it. Worth deciding whether the setting means "floor" or "base" before it gains more consumers.

Sirajmx:
@atc964 All three confirmed directly against the code before fixing — this was the most serious feedback of the batch, appreciated the thoroughness.

Pushed `c0ba011` addressing all three required items:

1. **Prune gating**: a real ad-server failure now sets `state.ad_server_sync_failed` and returns immediately — no mock fallback, no `_finish_sync()` call at all, so nothing in storage is touched on a transient outage. `_run_sync` and `POST /packages/sync` both check this flag and report `status: "error"` instead of false success; `_last_sync`/`_sync_count` no longer advance either. Added a test that seeds a real synced package, simulates a failing client, and asserts the package survives and the reported status is "error" — reverted the fix and confirmed both assertions fail as expected (the old code deleted the real package and reported "success").
2. **Floor default**: `default_price_floor_cpm` raised from `5.0` back to `10.0`, matching the literal it replaced. Added a test pinning the Settings class default directly so this can't silently drift again.
3. **Archived-item filter**: restored `status:ACTIVE` for GAM/FreeWheel, explicitly `None` for CSV (confirmed its `filter_str` is a literal substring match that would zero out results, exactly as you described). Added a test with a recording fake client confirming each ad-server type gets the filter it should.

Also applied the same `ad_server_sync_failed` honesty fix to `POST /packages/sync` for consistency, since it has the identical false-success shape.

Full suite: 1571 passed, 28 skipped. Ready for another look.

atc964 (dismissed):
All three required items check out. The failure path now sets `ad_server_sync_failed` and returns before `_finish_sync`, and `_prune_stale_synced_packages` has no other call site, so an outage cannot touch the synced layer. Both `_run_sync` and `POST /packages/sync` report status "error" on that path, and the scheduler no longer advances `_last_sync` on failure. The `status:ACTIVE` conditional is correct for CSV, and s3 is covered too, since `S3CsvAdServerClient` reuses `AdServerType.CSV` and its `list_inventory` ignores `filter_str` anyway. The committed failure-path test, seeded package survives a `ConnectionError`, is exactly what we asked for.

One leftover on AI-8 before this fully lands: `.env.example` line 139 still ships `DEFAULT_PRICE_FLOOR_CPM=5.0`, and `docs/guides/configuration.md` (lines 169 and 303) still documents the default as 5.0. Anyone copying the template explicitly reconfigures the exact halved floor this PR exists to remove. Please update both to 10.0 in this PR; it is a two-line change squarely in scope.

Two non-blocking notes, fine to track separately: (1) `POST /packages/sync` returns HTTP 200 with body status "error"; consider a 502/503 so REST clients checking status codes see the failure. (2) `include_archived` is now silently ignored; `admin.py` still passes `settings.inventory_sync_include_archived`, so an operator who set that flag loses archived inventory without warning. The docstring says tracked separately; please link the tracking issue on this PR.

Sirajmx:
@atc964 Pushed the 2-line fix (`abf27ad`) — `.env.example` and `docs/guides/configuration.md` (both the table entry and the example block) now read `DEFAULT_PRICE_FLOOR_CPM=10.0`, matching the code default. Ready to merge per your note.

